PR #6, opened 2024-06-03
Setup basic function to connect to the rpc consensus

marc-aurele-besner:
## Setup basic function to connect to the rpc consensus

- [x] Define the networks in a constants file (and types)
- [x] Define default network
- [x] Make a function that returns the network details from its id
- [x] Make a function that connects to a specific network of default network
- [x] Make functions that return network details
- [x] Make simple tests for all the above

### Reproduce

```yarn test```

### DIFF
--- a/packages/auto-consensus/src/api.ts
+++ b/packages/auto-consensus/src/api.ts
@@ -1,0 +1,35 @@
+import { ApiPromise, WsProvider } from '@polkadot/api'
+import { defaultNetwork, networks } from './constants/network'
+
+export const getNetworkDetails = (id?: string) => {
+  // If no id is provided, return the default network
+  if (!id) return defaultNetwork
+
+  // Find the network with the provided id
+  const network = networks.find((network) => network.id === id)
+  if (!network) throw new Error(`Network with id ${id} not found`)
+
+  return network
+}
+
+export const getNetworkRpcUrls = (id?: string) => {
+  // Get the network details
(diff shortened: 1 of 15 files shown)


abhi3700 (approved):
LGTM

packages/auto-consensus/src/api.ts
jfrank-summit, line 1: I was thinking this type of code would go into a shared package, as `auto-id` will also need api connections.

packages/auto-consensus/src/network.ts
jfrank-summit, line 1: I think this could also be utilized across consensus and an domain specific package.

jfrank-summit (commented):
A couple of comments on structures.

packages/auto-consensus/src/api.ts
marc-aurele-besner: Sounds good, that been said, I'm guessing I'll need to change my types a bit, since auto-id will likely use a rpc url specific to the id domain?

packages/auto-consensus/src/network.ts
marc-aurele-besner: (same comment as above)

packages/auto-consensus/src/api.ts
jfrank-summit: I think any likely shared code should be in a shared package.

jfrank-summit (approved):
LGTM

PR #10, opened 2024-06-04
Add dynamic save for both client and node

marc-aurele-besner:
## Add dynamic save for both client and node

- Add a generic dynamic save function that both support local storage or file system (use the key as the fileName + extension in the case of a file system


abhi3700 (dismissed):
LGTM

jfrank-summit (approved):

PR #11, opened 2024-06-04
Add a scripts folder in root with bash script that can run a local testnet

marc-aurele-besner:
## Add a scripts folder in root with bash script that can run a local testnet

Hopefully, this will make the development easier, especially as we start tackling extrinsics

### Localhost testing

To test the packages against a local node, you can use the script at `scripts/localhost.sh`.

1. Verify that the line 3-7 of the script matches your current OS and architecture.

   ```bash
   # Change the following variables as needed
   # OS to download
   OS="macos" # macos | ubuntu | windows
   # Architecture to download
   ARCHITECTURE="aarch64" # aarch64 | x86_64-skylake | x86_64-v2
   ```

2. Run the script:

   ```bash
   ./scripts/localhost.sh
   ```

   This script will download the latest version of the node and farmer for your OS and architecture, start the node, and farmer

3. Run the tests:

   ```bash
   bash scripts/localhost-run-test.sh
   ```

   The tests will detect the local node and farmer and run the tests against them instead of the public testnet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,7 @@ tsconfig.tsbuildinfo
 
 # Environment variables
 .env
+
+# Directory use by scripts
+executables/
+downloads/
--- a/README.md
+++ b/README.md
@@ -47,6 +47,36 @@ To run tests for all packages:
 
 `yarn test`
 
+### Localhost testing
+
+To test the packages against a local node, you can use the script at `scripts/localhost.sh`.
+
+1. Verify that the line 3-7 of the script matches your current OS and architecture.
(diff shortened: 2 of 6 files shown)


jfrank-summit (approved):
Looks great! This will work well for consensus, we may want to add support for auto-id in the future as well.

marc-aurele-besner:
> Looks great! This will work well for consensus, we may want to add support for auto-id in the future as well.

Indeed, I was just not to sure how to go about run an operator in localhost :)

PR #14, opened 2024-06-05
Switch from yarn to bun

marc-aurele-besner:
## Switch from yarn to bun

Yarn is a bit of a pain, the traditional yarn does not support import path such as workspace:*.
So I had first switched to yarn berry (canary -> 4.2.5) but it seems to be very inconsistent with there lock file, now using both yarn.lock and a couple .pnp files 🤯

And this seems to cause some weird incompatibility with vscode, as prettier on save and vs code does not detect the dependencies as installed anymore...

I tested bun last night, and it seems to work as we need-it out of the box and be faster


abhi3700 (approved):
LGTM

jfrank-summit:
@marc-aurele-besner I haven't used Bun but have heard some positive things about it. Are there any risks associated with a switch?

marc-aurele-besner:
> @marc-aurele-besner I haven't used Bun but have heard some positive things about it. Are there any risks associated with a switch?

That early on in the project, not really, the switch will have been a bit more painful if we had a super developed and complex mono repo.

The end user is still free to use Bun, Yarn, or NPM when integrating the packages. It's primarily for development and contribution that it matter.
Unlike a switch to Deno for example that make things, not backward compatible.

jfrank-summit (approved):

marc-aurele-besner:
After syncing with main, following #9 we have some issues that I can't solve.
Atm, I guess that some of the node:crypto implementations in Bun are either missing or not the same as on Node...
auto-id tests result with Bun

PR #21, opened 2024-06-10
Add utils to `auto-utils`, `auto-id` packages.

### LINKED ISSUE #15: Add `auto-kit`'s utils functions into auto-utils package.
This issue adds these functions:
- `der_encode_signature_algorithm_oid`
- `blake2b_256`

abhi3700:
Fixes #15 

This PR is to add the functions, classes from `auto-kit`'s [`utils.py`](https://github.com/subspace/auto-kit/blob/558666033fa0722a538da1be420c1555eed35e50/auto_identity/utils.py), [test_utils.py](https://github.com/subspace/auto-kit/blob/558666033fa0722a538da1be420c1555eed35e50/tests/auto_identity_tests/test_utils.py) (written in Python) to TypeScript code.

Added:
- `AlgorithmIdentifier`, `derEncodeSignatureAlgorithmOID` to `auto-id` package.
- `blake2b_256` to `auto-utils` package.
- corresponding tests as well.


packages/auto-id/package.json
```diff
@@ -10,7 +10,10 @@
   },
   "dependencies": {
     "@autonomys/auto-utils": "workspace:*",
-    "@types/node": "^20.12.12"
+    "@peculiar/asn1-schema": "^2.3.8",
+    "@peculiar/asn1-x509": "^2.3.8",
+    "@types/node": "^20.12.12",
+    "asn1-ts": "^8.0.2"
```
marc-aurele-besner, line 16: Is there only this package that does what we need?
It had a very low amount of download, and does not seem to be actively maintain...

packages/auto-id/package.json
```diff
@@ -10,7 +10,10 @@
   },
   "dependencies": {
     "@autonomys/auto-utils": "workspace:*",
-    "@types/node": "^20.12.12"
+    "@peculiar/asn1-schema": "^2.3.8",
+    "@peculiar/asn1-x509": "^2.3.8",
+    "@types/node": "^20.12.12",
```
marc-aurele-besner, line 15: Dependencies that only include types (so anything `@types/*`) should be in the `devDependencies` since they are only required at compilation time.

marc-aurele-besner (commented):
Looks good, with just a couple comments

packages/auto-id/package.json
abhi3700: Yes I have matched with the requirement of `auto-kit` in python.

> It had a very low amount of download, and does not seem to be actively maintain...

We are using the X509 certificate from there. Didn't find any library that's regularly updated. Also, these algorithms are generally not updated frequently. They tend to be more stable. Hence, less/no update for long time.

marc-aurele-besner: Yeah, it's not so much the low amount of recent commit that annoy me, more to see multiple Dependabot PR from many months ago, not merged and the low amount of downloads tbh.

abhi3700: At this moment, we don't have any alternative that I can find. If you are able to find a better lib, please do share the package.

marc-aurele-besner: What about https://www.npmjs.com/package/asn1js is it missing stuff?

abhi3700: I went for a TS variant.
- https://www.npmjs.com/package/asn1js: last commit 2 years ago.
- https://www.npmjs.com/package/asn1-ts: 9 months ago.

marc-aurele-besner: Regardless of it's name asn1js is fully type...
image
The repo has much more recent commit, they simply did not roll out recent version, we are still comparing a package with 1k or less download a week vs a package with 3.2M download a week...

PR #26, opened 2024-06-13
Add staking functions and tests

marc-aurele-besner:
## Add staking functions and tests

This PR adds staking functionality to the SDK + tests for them

***unlockFunds tests are missing as currently the 2 days lock make testing impossible***

### DIFF
--- a/packages/auto-consensus/__test__/staking.test.ts
+++ b/packages/auto-consensus/__test__/staking.test.ts
@@ -2,9 +2,16 @@ import { ActivateWalletInput, activateWallet } from '@autonomys/auto-utils'
 import { mnemonicGenerate } from '@polkadot/util-crypto'
 import { address } from '../src/address'
 import { balance } from '../src/balances'
-import { deregisterOperator, nominateOperator, registerOperator } from '../src/staking'
+import {
+  deregisterOperator,
+  nominateOperator,
+  registerOperator,
+  unlockFunds,
+  withdrawStake,
+} from '../src/staking'
 import { transfer } from '../src/transfer'
 import {
+  events,
   setup,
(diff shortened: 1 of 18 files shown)


packages/auto-consensus/__test__/staking.test.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    // Still unsure if this is testable due to the lock period of 2 days
+    // describe('Test unlockFunds()', () => {
+    //   test('Check Operator can unlockFunds after registration', async () => {
+    //     const mnemonicOperator = mnemonicGenerate()
+    //     const { api, accounts } = await activateWallet({
+    //       ...TEST_NETWORK,
+    //       uri: ALICE_URI,
+    //     } as ActivateWalletInput)
+    //     const { accounts: operatorAccounts } = await activateWallet({
```
abhi3700, line 260: Except for these commented tests, LGTM.

marc-aurele-besner:
> LGTM except for the commented tests which can be pushed in a separate PR or that's completed in this PR.

Yes so like I mention in the PR, atm these functions are not really testable due to the lock time hardcoded.
So I wonder if it's better to simply remove the tests, or if there is plan to make it a value that can be control via sudo, then the test could be used...

abhi3700:
> > LGTM except for the commented tests which can be pushed in a separate PR or that's completed in this PR.
> 
> Yes so like I mention in the PR, atm these functions are not really testable due to the lock time hardcoded. So I wonder if it's better to simply remove the tests, or if there is plan to make it a value that can be control via sudo, then the test could be used...

IMO, let's remove this then & add to a new PR.

PR #47, opened 2024-06-27
Add Registry & its registration example

abhi3700:
This PR resolves https://github.com/subspace/auto-sdk/issues/17


packages/auto-id/src/utils.ts
```diff
@@ -1,28 +1,73 @@
...
 
+// CLEANUP: Remove later when all registry functionalities work.
+// /**
+//  * Encodes a given string representation of an OID into its DER format.
+//  * This function is specifically used to encode signature algorithm OIDs.
+//  *
+//  * @param oid The string representation of the ObjectIdentifier to be encoded.
+//  * @returns Uint8Array containing the DER encoded OID along with NULL params of X.509 signature algorithm.
+//  */
+// export function derEncodeSignatureAlgorithmOID(oid: string): Uint8Array {
```
marc-aurele-besner, line 56: is this commented function still needed, since I see you changed the one above with the same name?

packages/auto-id/examples/registerAutoId.ts
marc-aurele-besner, line 1: can we keep the examples in the root directory?

marc-aurele-besner (commented):
just a few comments

packages/auto-id/examples/registerAutoId.ts
abhi3700: Yeah we would. But for now it's kept temporarily there. Once all the functions are completed. I will shift to the root folder for sure.
Need to organize it properly as to should there be only 1 example or more files.

packages/auto-id/src/utils.ts
abhi3700: yeah

packages/auto-id/examples/registerAutoId.ts
jfrank-summit: What is the benefit of keeping it here, even temporarily?

abhi3700: Nothing as such though. Was just thinking to put the final version into the root.
Anyway, I will put this to the root dir.

examples/auto-id/src/registerAutoId.ts
marc-aurele-besner, line 1: I'm not sure that I see the need for a separate example project for effectively one file.

Personally, I'll suggest you move `examples/auto-id/src/registerAutoId.ts` to `examples/node/src/regifesterAutoId.ts`
and modify `examples/node/package.json` scripts and dependencies section...

marc-aurele-besner: I think unifying the usage of auto-id and auto-consensus in same example project is better and easier to maintain 👍

abhi3700: I think then, IMO we should rename the folder from "node" to something else.
Moreover, more examples are coming in relation to auto-id.

packages/auto-id/src/utils.ts
```diff
@@ -1,28 +1,73 @@
...
+  // Combine the sequence header, OID, and NULL parameter into one Uint8Array
+  const derSequence = new Uint8Array(
+    sequenceHeader.length + oidEncoded.byteLength + nullParameter.length,
+  )
+  derSequence.set(sequenceHeader, 0)
+  derSequence.set(new Uint8Array(oidEncoded), sequenceHeader.length)
+  derSequence.set(nullParameter, sequenceHeader.length + oidEncoded.byteLength)
+
+  // Apply SCALE encoding by prepending the compact length of the entire DER sequence
+  return compactAddLength(derSequence)
```
jfrank-summit, line 45: Is applying SCALE encoding really necessary? I don't recall doing this in `auto-kit`.

packages/auto-id/src/utils.ts
```diff
@@ -1,28 +1,73 @@
...
+  }
 
-  return new Uint8Array([...sequenceHeader, ...new Uint8Array(berArrayBuffer), ...nullParameter])
+  // Combine the sequence header, OID, and NULL parameter into one Uint8Array
+  const derSequence = new Uint8Array(
+    sequenceHeader.length + oidEncoded.byteLength + nullParameter.length,
+  )
+  derSequence.set(sequenceHeader, 0)
+  derSequence.set(new Uint8Array(oidEncoded), sequenceHeader.length)
+  derSequence.set(nullParameter, sequenceHeader.length + oidEncoded.byteLength)
```
jfrank-summit, line 42: This seems like a quite manual way to handle the encoding. Does the asn1js or other libraries not handle this for you? Did you look into [@peculiar/asn1-schema](https://www.npmjs.com/package/@peculiar/asn1-schema)?

jfrank-summit (commented):
A couple of comments.

packages/auto-id/src/utils.ts
abhi3700: Yes in TS. Otherwise, it's not taking the argument properly.
I had created an issue on slack: https://subspacenetwork.slack.com/archives/C03QY5Z0USF/p1719385961803199 before using SCALE.

packages/auto-id/src/utils.ts
abhi3700: As far as I can recall, there was issue in bytes count which on debugging I found the current code as solution as a fix.

> Did you look into [@peculiar/asn1-schema](https://www.npmjs.com/package/@peculiar/asn1-schema)

Yes I had, but didn't found anything relevant.

packages/auto-id/src/utils.ts
jfrank-summit: The SCALE encoding should not happen here. If it is needed to submit the transaction, it should happen in the submission code. This should just return the DER encoded signature algorithm.

packages/auto-id/src/utils.ts
jfrank-summit: This code looks very fragile. Did you look at the [AlgorithmIdentifier class](https://github.com/PeculiarVentures/asn1-schema/blob/c800e1286d52b460a63f665f90d8e74d6ac69067/packages/x509/src/algorithm_identifier.ts). This looks similar to what was used in [auto-kit](https://github.com/subspace/auto-kit/blob/558666033fa0722a538da1be420c1555eed35e50/auto_identity/utils.py#L7)

packages/auto-id/src/utils.ts
abhi3700: I get this error, if i don't use SCALE encoding: `compactAddLength(derSequence)`.
```sh
Error: createType(Call):: Call: failed decoding autoId.registerAutoId:: Struct: failed on args: {"req":"{\"_enum\":{\"X509\":\"PalletAutoIdRegisterAutoIdX509\"}}"}:: Struct: failed on req: {"_enum":{"X509":"PalletAutoIdRegisterAutoIdX509"}}:: Enum(X509):: Enum(Root):: Struct: failed on signatureAlgorithm: Bytes:: Bytes: required length less than remainder, expected at least 13, found 9
at createTypeUnsafe (/Users/abhi3700/F/coding/github_repos/subspace/auto-sdk/node_modules/@polkadot/types-create/cjs/create/type.js:54:22)
    at TypeRegistry.createTypeUnsafe (/Users/abhi3700/F/coding/github_repos/subspace/auto-sdk/node_modules/@polkadot/types/cjs/create/registry.js:230:52)
    at extrinsicFn (/Users/abhi3700/F/coding/github_repos/subspace/auto-sdk/node_modules/@polkadot/types/cjs/metadata/decorate/extrinsics/createUnchecked.js:16:25)
    at Object.decorated [as registerAutoId] (/Users/abhi3700/F/coding/github_repos/subspace/auto-sdk/node_modules/@polkadot/api/cjs/base/Decorate.js:491:50)
    at receipt (/Users/abhi3700/F/coding/github_repos/subspace/auto-sdk/packages/auto-id/src/registry.ts:166:26)
    at new Promise (<anonymous>)
    at Registry.<anonymous> (/Users/abhi3700/F/coding/github_repos/subspace/auto-sdk/packages/auto-id/src/registry.ts:165:46)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/abhi3700/F/coding/github_repos/subspace/auto-sdk/packages/auto-id/src/registry.ts:28:58)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

jfrank-summit: Per the name (`derEncodeSignatureAlgorithmOID`) this function should return DER encoded data. I think you should SCALE encode it [here](https://github.com/subspace/auto-sdk/blob/e88852cc620da5c7b9058e591e1c66e38bb77855/packages/auto-id/src/registry.ts#L107))

marc-aurele-besner (approved):
I just trigger the action that had failed again and this time it worked...

jfrank-summit (approved):

PR #53, opened 2024-07-03
Create a cli to target all functionality

marc-aurele-besner:
## Create a cli to target all functionality

This adds a crude CLI in `examples/node/`

```bash
cd examples/node
yarn cli
```

or
```bash
yarn cli transfer
```

It seems to be bugging on function with a transaction, I'm not sure if this is just terminal display bugs or else...

Next step will be to pass some parameters to pass values to the underlying function

### Screenshots

auto-sdk


jfrank-summit (approved):
lgtm

PR #55, opened 2024-07-03
Publish workflow improvement and next example improvements

marc-aurele-besner:
## Publish workflow improvement and next example improvements

The main reason the workflow was failing was `.yarn/install-state.gz` will change despite been in .gitignore
I deleted the file and now it seems to not been tracked anymore...

But when reproducing the step

`yarn publish` still require a human interaction to chose the versioning, so we will need to default to a set versioning bump if we want the publishing flow automated.

Also a few improvement in the next js example


jfrank-summit:
I feel like I am missing context here. The changes don't seem to align with the description.

marc-aurele-besner:
> I feel like I am missing context here. The changes don't seem to align with the description.

Yeah I did fix/improves a few things in the next js example also in the same pr 🤦‍♂️

jfrank-summit (dismissed):

PR #67, opened 2024-07-11
Improve Auto-Utils

marc-aurele-besner:
## Improve Auto-Utils

- Hide init warning on connection to API
- Improce code redundance
- Allow to pass a api as props in activateWallet() and mockWallets()


packages/auto-utils/src/api.ts
```diff
@@ -1,29 +1,40 @@
...
+    noInitWarn: options?.noInitWarn ?? true,
+    provider,
+  })
   await api.isReady
 
   return api
 }
 
-export const activateDomain = async (input: DomainInput): Promise<ApiPromise> => {
+export const activate = async (input?: ActivateInput<NetworkInput>): Promise<ApiPromise> => {
```
jfrank-summit, line 22: It is not very clear to me what `input` represents? Is it options or params? The naming feels off here, but I may just not be following.

jfrank-summit:
Tests are timing out.

packages/auto-utils/src/api.ts
marc-aurele-besner: yeah I could replace there by options or params...

it take in a optional networkId (default to 3h if nothing is pass) and a optional api, so if you activate multiple wallet you can initialize 1 api and use the same for all wallet to save time

jfrank-summit (approved):

PR #73, opened 2024-07-15
Remove unnecessary async and awaits

jfrank-summit:
There were some unnecessary async and awaits on extrinsics.


marc-aurele-besner (approved):
LGTM 🚀

PR #103, opened 2024-09-12
feat: add auto-drive package

clostao:
This PR implements these features:

- CID management: How is calculated the CID of a IPLD node. We use multihash with base32 encoding and `blake3` as the hash function
- Metadata management: Both offchain and onchain metadata. Offchain metadata is more extended as it doesn't have same length limitations as onchain metadata which is much more short.
- IPLD Dag Creation: How a file is translated into a IPLD DAG.


clostao:
@jfrank-summit

packages/auto-drive/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "typescript": "^5.6.2"
+  },
+  "dependencies": {
+    "@ipld/dag-pb": "^4.1.2",
+    "blake3": "1.1.0",
+    "multiformats": "^13.2.2",
+    "protobufjs": "^7.4.0",
+    "protons": "^7.6.0",
+    "protons-runtime": "^5.5.0"
+  }
```
marc-aurele-besner, line 32:
```suggestion
  },
  "repository": {
    "type": "git",
    "url": "https://github.com/autonomys/auto-sdk"
  },
  "author": {
    "name": "Autonomys",
    "url": "https://www.autonomys.net"
  }
```

packages/auto-drive/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './cid/index.js'
```
marc-aurele-besner, line 1: I believe you could remove all the .js

marc-aurele-besner (commented):
Awesome, LGTM, just a couple comments 👍

packages/auto-drive/src/index.ts
clostao: I tried, but since I'm using ES modules due to the dependencies of the package. `moduleResolution` is forced to `NodeNext` and it requires having extension.

marc-aurele-besner (approved):
LGTM 🚀

PR #117, opened 2024-10-04
Add metadata dag builder 

clostao:
Added metadata dag builder abstracting the node building through `builders` parameter


marc-aurele-besner (changes requested):
Can you also add `createMetadataIPLDDag` to the Auto-Drive README?

marc-aurele-besner (approved):
Awesome

PR #131, opened 2024-10-22
Enable Processing of Large Files to IPLD Format

clostao:
This PR solves one of the current limitations of AutoDrive's file upload limit: NodeJS Buffer object allows up to byte-length of 1GB

For solving this issue, this PR substitutes the parameter of `createBufferIPLDDag` Buffer for a AsyncIterable<Buffer>. Now the limit of an upload size depends on the user's RAM (among other factors) rather a fixed amount by NodeJS (1Gb).


packages/auto-drive/src/ipld/chunker.ts
```diff
@@ -15,57 +15,69 @@ export interface IPLDDag {
...
-  file: Buffer,
+  file: AsyncIterable<Buffer>,
   filename?: string,
   { chunkSize, maxLinkPerNode }: { chunkSize: number; maxLinkPerNode: number } = {
     chunkSize: DEFAULT_MAX_CHUNK_SIZE,
     maxLinkPerNode: DEFAULT_MAX_LINK_PER_NODE,
   },
-): IPLDDag => {
+): Promise<IPLDDag> => {
   return createBufferIPLDDag(file, filename, fileBuilders, { chunkSize, maxLinkPerNode })
```
marc-aurele-besner, line 25: If createBufferIPLDDag() is async, then we are likely missing an `await`

packages/auto-drive/src/ipld/chunker.ts
```diff
@@ -15,57 +15,69 @@ export interface IPLDDag {
...
   limits: { chunkSize: number; maxLinkPerNode: number } = {
     chunkSize: DEFAULT_MAX_CHUNK_SIZE,
     maxLinkPerNode: DEFAULT_MAX_LINK_PER_NODE,
   },
-): IPLDDag => {
+): Promise<IPLDDag> => {
   const buffer = Buffer.from(JSON.stringify(metadata))
   const name = `${metadata.name}.metadata.json`
-  return createBufferIPLDDag(buffer, name, metadataBuilders, limits)
+  return createBufferIPLDDag(
```
marc-aurele-besner, line 37: If createBufferIPLDDag() is async, then we are likely missing an `await`

marc-aurele-besner (dismissed):
LGTM

marc-aurele-besner (approved):
LGTM

marc-aurele-besner (approved):
LGTM

clostao:
A little update about this PR: 

What I got previously reviewed was to enable the user to digest into the formatter over the MAX_BUFFER_SIZE of NodeJS (1gb-2gb). 

The changes after the review are to solve the issue of with the IPLDDag return type structure. With that structure the user of these methods had to get all the file (formatted in IPLD) in memory, imposible for files in few tens of GBs. For solving this I decided to use a solution that [ipfs-unixfs-importer](https://github.com/ipfs/js-ipfs-unixfs/tree/main/packages/ipfs-unixfs-importer) already uses for solving this very same issue. Basically, using an abstract interface of `Blockstore` enables the user to manage the storage of IPLD nodes through the best fit for their solution. e.g For auto-drive application we'll use a postgres-based Blockstore avoiding the requirement of having the uploading files in memory.

packages/auto-drive/package.json
```diff
@@ -34,6 +34,9 @@
   "dependencies": {
     "@ipld/dag-pb": "^4.1.2",
     "blake3": "1.1.0",
+    "blockstore-core": "^5.0.2",
+    "interface-blockstore": "^5.3.1",
```
marc-aurele-besner, line 38: `interface-blockstore` does not seem to be use in your PR, is it a peerDependencies?

packages/auto-drive/package.json
```diff
@@ -34,6 +34,9 @@
...
+    "interface-store": "^6.0.2",
```
marc-aurele-besner, line 39: `interface-store` seem to only be use for types, if so, it could be moved to the devDependencies

packages/auto-drive/src/ipld/chunker.ts
```diff
@@ -1,103 +1,138 @@
 import { PBNode } from '@ipld/dag-pb'
+import { BaseBlockstore } from 'blockstore-core'
+import { AwaitIterable } from 'interface-store'
```
marc-aurele-besner, line 3:
```suggestion
import type { AwaitIterable } from 'interface-store'
```

packages/auto-drive/src/ipld/chunker.ts
```diff
@@ -1,103 +1,138 @@
...
-import { cidOfNode } from '../cid/index.js'
-import { OffchainMetadata } from '../metadata/index.js'
+import { cidOfNode, cidToString } from '../cid/index.js'
+import { decodeIPLDNodeData, IPLDNodeData, OffchainMetadata } from '../metadata/index.js'
 import { Builders, fileBuilders, metadataBuilders } from './builders.js'
 import { createFolderInlinkIpldNode, createFolderIpldNode } from './nodes.js'
 import { chunkBuffer, encodeNode } from './utils.js'
 
 export const DEFAULT_MAX_CHUNK_SIZE = 1024 * 64
 export const DEFAULT_MAX_LINK_PER_NODE = DEFAULT_MAX_CHUNK_SIZE / 64
```
marc-aurele-besner, line 12: It seems weird to first multiply 1024 by 64, then dividing it by 64, Why not switching the order
```
export const DEFAULT_MAX_LINK_PER_NODE = 1024
export const DEFAULT_MAX_CHUNK_SIZE = DEFAULT_MAX_LINK_PER_NODE * 64
```

packages/auto-drive/tests/chunker.spec.ts
```diff
@@ -1,17 +1,26 @@
-import { createNode } from '@ipld/dag-pb'
-import { cidOfNode } from '../src'
-import { createFileIPLDDag, createFolderIPLDDag, createMetadataIPLDDag } from '../src/ipld/chunker'
+import { createNode, decode, PBNode } from '@ipld/dag-pb'
```
marc-aurele-besner, line 1: I see in your test you still use decode and other function from @ipld/* dependencies, could we expose the functions a user will need from our package, so they can import everything they need, directly from our package, and apply the same behavior in our test?

marc-aurele-besner:
@clostao I had already given my review 😉

clostao:
My bad

packages/auto-drive/package.json
clostao: Thanks for the catch!

packages/auto-drive/src/ipld/chunker.ts
clostao: I agree is weird, I updated to be less confusing though I'd prefer to show `DEFAULT_MAX_LINK_PER_NODE` dependant on `DEFAULT_MAX_LINK_PER_NODE` because shows better the logical process that I made for those numbers.

marc-aurele-besner (approved):
LGTM

jfrank-summit (approved):
Is there no limit on upload size with this update @clostao?

marc-aurele-besner:
I'm guessing the limit become the practical cost of the set of transactions require uploading all these cid's and data chunk

clostao:
> Is there no limit on upload size with this update @clostao?

Theoretically no, it's true that this is implementation is the first iteration of asynchronous IPLD node building so maybe not to optimised and for uploading hundreds-GB sized or TB-sized files may be slow and somehow inconvenient yet. 

We'll know better once it's deployed this PR in the server because the difference between local testing and deployed in processes this sensitive to server latency. I tested with 2GB files and the processing of the file happened within <9s. 

That being said this are numbers roughly tested and I'd prefer not to stick my neck out but I'm confident that this will enable the upload of tens of GBs and if not with optimisations of this approach.

marc-aurele-besner:
Also, this is the SDK, so I imagine we will keep a conservative limit on the auto-drive product
This is not stopping anyone to run the SDK locally, to prepare all the CID and chunks and upload themselves to the network

clostao:
> Also, this is the SDK, so I imagine we will keep a conservative limit on the auto-drive product This is not stopping anyone to run the SDK locally, to prepare all the CID and chunks and upload themselves to the network (therefore not having much latency to prepare the CID)

You're right, here latency is no issue here.

PR #144, opened 2024-10-31
Update pkg name and added encryption & compression logic

clostao:
* Updates the package name from `@autonomys/auto-drive` to `@autonomys/auto-dag-data`
* Implements the encryption/compression mechanism implemented in [@autonomys/auto-drive](https://github.com/autonomys/auto-drive/pull/72) with a generalized interface for future algorithm additions


packages/auto-dag-data/README.md
```diff
@@ -1,14 +1,14 @@
-# Auto-Drive
+# Auto Dag Data
```
jfrank-summit, line 1:
```suggestion
# Auto DAG Data
```

packages/auto-dag-data/src/encryption/index.ts
```diff
@@ -0,0 +1,47 @@
...
+import { asyncByChunk } from '../utils/async.js'
+
+const crypto = new Crypto()
+
+export const ENCRYPTING_CHUNK_SIZE = 1024 * 1024
+const IV_SIZE = 16
+const TAG_SIZE = 16
+const ENCRYPTED_CHUNK_SIZE = ENCRYPTING_CHUNK_SIZE + IV_SIZE + TAG_SIZE
+
+const getKeyFromPassword = async (password: string) => {
```
jfrank-summit, line 20: Is there a reason to not salt the password?

clostao: Not a good one, I just copied from the PoC. We should enable salting.

clostao: Thanks for the catch

clostao: Done

clostao:
I'd have to get this merged to publish the package `@autonomys/auto-dag-data` since `@autonomys/auto-drive` is dependant to this one

marc-aurele-besner (approved):

packages/auto-dag-data/src/encryption/index.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const getKeyFromPassword = async ({ password, salt }: PasswordGenerationOptions) => {
+  const encoder = new TextEncoder()
+  const saltHash =
+    typeof salt === 'string' ? await crypto.subtle.digest('SHA-256', encoder.encode(salt)) : salt
+
+  const passwordPlusSalt = Buffer.concat([Buffer.from(saltHash), encoder.encode(password)])
+  const passwordHash = await crypto.subtle.digest('SHA-256', passwordPlusSalt)
+
+  return await crypto.subtle.importKey('raw', passwordHash, { name: 'AES-GCM' }, false, [
```
jfrank-summit, line 20: I think it would be better to use `deriveKey` here with with [PBKDF2](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveKey#supported_algorithms). Something like:
```
const getKeyFromPassword = async ({ password, salt }: PasswordGenerationOptions) => {
  const encoder = new TextEncoder()
  const saltHash = typeof salt === 'string' 
    ? await crypto.subtle.digest('SHA-256', encoder.encode(salt)) 
    : salt

  // First import password as key material
  const keyMaterial = await crypto.subtle.importKey(
    'raw',
    encoder.encode(password),
    'PBKDF2',
    false,
    ['deriveBits', 'deriveKey']
  )

  // Then use PBKDF2 to derive the actual key
  return await crypto.subtle.deriveKey(
    {
      name: 'PBKDF2',
      salt: saltHash,
      iterations: 100000,
      hash: 'SHA-256'
    },
    keyMaterial,
    { name: 'AES-GCM', length: 256 },
    false,
    ['encrypt', 'decrypt']
  )
}
```

jfrank-summit (commented):
I had one comment on our encryption key derivation approach. Also, could you add tests for compression and encryption. Thanks!

packages/auto-dag-data/src/encryption/index.ts
clostao: Yeah, looks much better

clostao: Aren't 100_000 iterations way too much?

clostao: Nvm, I just look for it and looks like they aren't too much kind of surprising though

packages/auto-dag-data/src/encryption/index.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  const key = await getKeyFromPassword(passwordGenerationOptions)
+
+  for await (const chunk of asyncByChunk(file, chunkSize)) {
+    const iv = chunk.subarray(0, IV_SIZE)
+    const encryptedChunk = chunk.subarray(IV_SIZE, chunkSize)
+    const decrypted = await crypto.subtle.decrypt({ name: 'AES-GCM', iv }, key, encryptedChunk)
+    yield Buffer.from(decrypted)
+  }
```
jfrank-summit, line 74: How are you currently handling the storing of the salt? I believe you can prepend it to the encrypted file and then extract it during decryption and then not have to store it separately. This will allow you to generate the salt randomly during file encryption.

clostao: I was thinking of having a static salt per service but prepending is much better solution. I just implemented it. Thanks!

jfrank-summit (approved):
LGTM.

PR #160, opened 2024-11-08
Add build in publish script

marc-aurele-besner:
## Add build in publish script


marc-aurele-besner:
I had look at your last pr very quickly, not realizing you had implemented a very similar solution

PR #194, opened 2024-12-04
Separate into node and browser outputs

clostao:
Splits`@autonomys/auto-drive` into two main files one used for browser environments and the other for server enviroments.

Implements [rollup](https://rollupjs.org) for doing so.


marc-aurele-besner (approved):
Awesome :)

marc-aurele-besner (approved):

PR #195, opened 2024-12-05
Add provider support for multiple endpoints

marc-aurele-besner:
## Add provider support for multiple endpoints

I haven't tested yet if this will automatically switch endpoint if one of them is not working


clostao (approved):
[Looks like should work fine](https://github.com/polkadot-js/api/blob/master/packages/rpc-provider/src/ws/index.ts#L195)

PR #199, opened 2024-12-10
Translate Auto-Drive Into CJS

clostao:
Updates auto-drive into a cjs module by. 

- Importing functions and classes asynchronously
- Duplicating some of the types in `@autonomys/auto-dag-data`
- Removed rollup since it won't be used for multiple output (`esm` and `cjs`) build


marc-aurele-besner (approved):
Awesome improvement to devEx!
Good job! 👏

PR #200, opened 2024-12-11
Bumped to v1.1.0

clostao:
- Updated version of v1.1.0
- Additionally I updated the examples since there was a build error


marc-aurele-besner (approved):
Awesome 🥳

PR #208, opened 2024-12-19
update: improve observable returning type

clostao:
Having the CID as optional was a weird experience. Adding distinctly typed by the field completed makes more comprehensible.

### DIFF
--- a/packages/auto-drive/src/api/models/uploads.ts
+++ b/packages/auto-drive/src/api/models/uploads.ts
@@ -74,33 +74,22 @@ export type CompleteUploadResponse = {
   cid: string
 }
 
-export type UploadFileStatus =
+type ObjectUploadStatus<Type extends 'file' | 'folder'> =
   | {
       completed: true
-      type: 'file'
+      type: Type
       progress: number
       cid: string
     }
   | {
       completed: false
-      type: 'file'
+      type: Type
       progress: number
       cid: null
     }
 
-export type UploadFolderStatus =
-  | {
-      completed: true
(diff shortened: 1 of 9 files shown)


packages/auto-drive/src/api/models/uploads.ts
```diff
@@ -74,17 +74,33 @@ export type CompleteUploadResponse = {
...
+      progress: number
+      cid: null
+    }
 
-export type UploadFolderStatus = {
-  type: 'folder'
-  progress: number
-  cid?: string
-}
+export type UploadFolderStatus =
```
marc-aurele-besner, line 91: Personally I find these two types a bit repetitive, why not making use of generic types?

```typescript
type UploadStatus<Type extends 'file' | 'folder'> = 
  | {
      completed: true
      type: Type
      progress: number
      cid: string
    }
  | {
      completed: false
      type: Type
      progress: number
      cid: null
    }

export type UploadFileStatus = UploadStatus<'file'>
export type UploadFolderStatus = UploadStatus<'folder'>
```

marc-aurele-besner (commented):
LGTM, that been said, I find these two almost identical types a bit redundant

packages/auto-drive/src/api/models/uploads.ts
clostao: Updated! Thanks

marc-aurele-besner (approved):
LGTM

clostao:
Closing because is outdated

PR #232, opened 2025-02-07
feat: add shareable objects

clostao:
Adds support for shareable objects. Implements `publishObject` which returns the link to the public downloadable file.


marc-aurele-besner (approved):
LGTM Thank you for updating the readme

PR #247, opened 2025-03-04
Fix transfers parsing

marc-aurele-besner:
## Fix transfers parsing

- Fix the method for balances sections
- Improve parsing of the signer
- Improve signature and args typing


clostao (approved):
LGTM

PR #288, opened 2025-03-17
Add async library

clostao:
A toolbox library for managing complex async problems such as concurrency or managin async generators

### DIFF
--- a/packages/auto-dag-data/tests/chunker.spec.ts
+++ b/packages/auto-dag-data/tests/chunker.spec.ts
@@ -46,7 +46,8 @@ describe('chunker', () => {
       /// Check the metadata
       expect(decoded.name).toBe(name)
       expect(decoded.type).toBe(MetadataType.File)
-      expect(Buffer.from(decoded.data ?? '').toString()).toBe(text)
+      const data = decoded.data ? Buffer.from(decoded.data) : Buffer.from('')
+      expect(data.toString()).toBe(text)
       expect(decoded.linkDepth).toBe(0)
       expect(decoded.size?.toString()).toBe(text.length.toString())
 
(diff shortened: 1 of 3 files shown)


marc-aurele-besner (changes requested):
I'm fine with the content of the package, but not with it's name. These functions are all very specific, to handle streams, concurrency and `async` seems too much of a catch all name.

When I read `async` this is what I have in mind https://www.npmjs.com/package/async

clostao:
I understand that the package name might be confused with the one you mentioned, and we definitely want to avoid that. It’s a bit tricky to choose a name that covers both generators (async function*) and concurrency without coming off as too generic about asynchronous processes.

So basically, we have two routes: we can either break the functionality into two separate packages (like @autonomys/generators and @autonomys/concurrency) or rename it to @autonomys/asynchronous to avoid the naming conflict. I personally lean more for the latter since it gives us more room to add features later, though I don't have a big issue with splitting them up.

marc-aurele-besner:
> I understand that the package name might be confused with the one you mentioned, and we definitely want to avoid that. It’s a bit tricky to choose a name that covers both generators (async function*) and concurrency without coming off as too generic about asynchronous processes.
> 
> So basically, we have two routes: we can either break the functionality into two separate packages (like @autonomys/generators and @autonomys/concurrency) or rename it to @autonomys/asynchronous to avoid the naming conflict. I personally lean more for the latter since it gives us more room to add features later, though I don't have a big issue with splitting them up.

I'm also fine with the latter

marc-aurele-besner:
Sorry, I was annoying of all these renovate PR's and wanted to get them in 😅

marc-aurele-besner (approved):

marc-aurele-besner (approved):

marc-aurele-besner (dismissed):
One of the test still seems to fail

marc-aurele-besner (approved):

PR #302, opened 2025-03-18
Published v1.4.1

clostao:
The standalone publication of the @autonomys/asynchronous v.1.4.0 was missing the compiled package.


marc-aurele-besner (approved):
Do we need the `prebuild` command in the root package.json?

clostao:
Nope, lerna is handling that dependency rn, I think

marc-aurele-besner (approved):

PR #311, opened 2025-03-24
Add custom-RPC package

clostao:
As we've talked this package would be filled with tools used for implementing custom RPC protocols like [object mapping indexers](https://github.com/autonomys/auto-files-gateway/blob/main/services/object-mapping-indexer/src/controllers/objects.ts#L151) or subspace gateway. 

This is the first version that implements: 

- Easy way to define RPC handlers with optional responses
- Handling reconnections in WS close/errors
- Basic typing in responses

Soon I'll implement better type definition: request-level type definition, response definition...


packages/rpc/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@autonomys/rpc",
+  "packageManager": "yarn@4.7.0",
+  "version": "1.4.1",
+  "private": true,
```
marc-aurele-besner, line 5: So this package won't be published?
On the other way, if it's going to be published, let's add the basic template we use for other package.json

packages/rpc/src/models/common.ts
```diff
@@ -0,0 +1,25 @@
+import z from 'zod'
```
marc-aurele-besner, line 1: `zod` is not in package.json

packages/rpc/src/rpc/client.ts
```diff
@@ -0,0 +1,86 @@
...
+  callbacks: {
+    onOpen?: () => void
+    onReconnection?: () => void
+    onError?: (error: Error) => void
+    onClose?: (event: Websocket.ICloseEvent) => void
+    onWrongMessage?: (responder: (message: string) => void) => void
+  }
+  reconnectInterval?: number | null
+}): ClientRPC => {
+  let ws: WsClient = createWsClient({
```
marc-aurele-besner, line 25: I don't see anywhere `ws` been reassigned, I believe it could be a const

marc-aurele-besner (commented):
Just a couple comments/questions, looking good 👍

marc-aurele-besner (approved):
LGTM

PR #314, opened 2025-03-25
File caching package

clostao:
Package for managing FS-based file caching with two main default configs:

- Single memory layer 
- Double memory-sqlite3 layer


marc-aurele-besner (approved):
LGTM

PR #321, opened 2025-03-25
Fix rpc response

clostao:
The controller for rpc responses expected a `string` where it should receive a `MessageResponseQuery`

Also I updated the type `MessageResponseQuery` for easing the devEx


marc-aurele-besner (approved):
LGTM

PR #323, opened 2025-03-26
Fixing/enhancing RPC server

clostao:
- RPC Callbacks were passing the whole message rather than the expect `params` field
- The server listening was a bit more tricky than expected so I added a new method in rpc for listening


marc-aurele-besner (approved):
LGTM

PR #326, opened 2025-03-27
feat: add auto-dag-data methods

clostao:
Wrap into async methods some useful tools in `@autonomys/auto-dag-data`

### DIFF
--- a/packages/auto-drive/package.json
+++ b/packages/auto-drive/package.json
@@ -42,9 +42,10 @@
     "typescript": "^5.6.3"
   },
   "dependencies": {
-    "@autonomys/asynchronous": "^1.4.6",
-    "@autonomys/auto-dag-data": "^1.4.6",
-    "@autonomys/auto-utils": "^1.4.6",
+    "@autonomys/asynchronous": "^1.4.5",
+    "@autonomys/auto-dag-data": "^1.4.5",
+    "@autonomys/auto-utils": "^1.4.5",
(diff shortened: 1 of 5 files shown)


packages/auto-drive/package.json
```diff
@@ -42,9 +42,10 @@
     "typescript": "^5.6.3"
   },
   "dependencies": {
-    "@autonomys/asynchronous": "^1.4.6",
-    "@autonomys/auto-dag-data": "^1.4.6",
-    "@autonomys/auto-utils": "^1.4.6",
+    "@autonomys/asynchronous": "^1.4.5",
```
marc-aurele-besner, line 45: look like main is already at 1.4.6...

packages/auto-drive/src/offline/cid.ts
```diff
@@ -0,0 +1,87 @@
...
+import { GenericFile } from '../api/models/file'
+
+const getFileAsync = async (file: File | GenericFile | Buffer) => {
+  if (file instanceof File) {
+    return await file.arrayBuffer()
+  }
+  return file
+}
+
+// export const precomputeCid = async (file: File | GenericFile | Buffer, name?: string) => {
```
marc-aurele-besner, line 11: why this commented function?

clostao:
Thanks for the catch! Should be fine now

marc-aurele-besner (approved):
LGTM

PR #327, opened 2025-03-27
Add RPC API definitions for implementing type-safe RPC protocols 

clostao:
This PR implements a system for defining RPC APIs for achieveing type safety in build time during implementation of custom RPC protocols.


packages/rpc/index.ts
```diff
@@ -0,0 +1,41 @@
...
+    params: defineUnvalidatedType<{ name: string }>(),
+    returns: z.object({
+      name: z.string(),
+    }),
+  },
+})
+
+server
+  .createClient({
+    endpoint: 'ws://localhost:8080',
```
marc-aurele-besner, line 18: is it normal that this endpoint is hardcoded?

clostao: I accidentally upload that one

marc-aurele-besner (approved):
LGTM

PR #332, opened 2025-03-28
Fix `returns` rpc api definition

clostao:
The issue is that `returns` expected to define the RPC message instead of what is thought for that is to define the format of the result of the RPC message.  

This led to error defining the APIs since the message & message result were incompatible types.

Other:

- I've added also an improvement is handled the absence of messageId
- Added *.debug.ts to `.gitignore` to avoid uploading debugging/local testing ts files


marc-aurele-besner (dismissed):
LGTM

marc-aurele-besner (approved):

PR #334, opened 2025-03-30
Feat: user session package

marc-aurele-besner:
## Feat: user session package

Add a package that has a simple Smart Contract to store user session, wrap Auto Drive and export the ABI and function to use the contract.


iamnamananand996 (approved):
Looks good.

Just curious to know, how it will be used in our application.

Do we add some example also ?

clostao (approved):

packages/user-session/src/utils/data.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!userSession || userSession === '0x') {
+      logger(params, 'User session not found')
+      return null
+    }
+
+    const cid = userSessionCIDFromHash(userSession)
+    logger(params, 'cid:', cid)
+
+    logger(params, `Downloading file: ${cid}`)
+    const stream = await autoDriveApi.downloadFile(cid, params.password)
```
clostao, line 27: Just some style comments. You can use [asyncIterableToBuffer](https://github.com/autonomys/auto-sdk/blob/52d4a5b3e5451d67fe3bf118d5360e9717b0c90a/packages/asynchronous/src/asyncGenerators/index.ts#L51) from `@autonomys/asynchronous` for transform into a buffer.

marc-aurele-besner: ah, nice!
Thank you, I'll merge this and add this improvement in the next version.
There is a few other improvements I have in mind, and I'm sure as I integrate it to the explorer and the faucet, I'll have other thing to improve on.

Thank you

PR #342, opened 2025-04-01
Add notifications support to Type-Safe RPC

clostao:
In addition to the Message (client request + server response) rpc feature server should be able to send messages to the client for enabling mechanism like subscription to events.

This wasn't support before and this PR enables server to send notifications


marc-aurele-besner (approved):
LGTM

marc-aurele-besner (approved):

PR #352, opened 2025-04-03
fix: caching policy

clostao:
Using nonBlocking policy was making that no file was stored in sqlite cache leading to cache loss when service reboots. 

Additionally, I was generating incorrectly the path for the file to be stored though it wasn't causing problems due to the partiotioning of folders.


jfrank-summit (approved):
lgtm

PR #355, opened 2025-04-04
v1.4.14

clostao:


clostao:
This https://github.com/autonomys/auto-sdk/pull/353 was merged after the publication of the package but the version update was merged afterwards

PR #364, opened 2025-04-08
Feat: Add WAI3 contract

marc-aurele-besner:
## Feat: Add WAI3 contract

Add a @autonomys/contracts package to put all our basic contracts that we plan on deploying on both testnet and mainnet, with a package that export the ABI for these and a contract wrapper


clostao (approved):
LGTM

PR #367, opened 2025-04-11
fix: temporary file path

clostao:
The issue is that when a CID is triggered twice before the first trigger cache is fully processed there's a conflict in the filepath of the temporary cache, resulting into failed rename operations since the temporary file doesn't exist due to it was renamed with the first trigger processing.


package.json
```diff
@@ -50,5 +50,8 @@
     "prettier": "^3.5.3",
     "ts-jest": "^29.3.0",
     "typescript": "^5.4.5"
+  },
+  "dependencies": {
+    "uuid": "^11.1.0"
```
marc-aurele-besner, line 55: this should be added to the respective package.json project not the root package.json

marc-aurele-besner (changes requested):

iamnamananand996 (approved):

marc-aurele-besner (approved):

PR #386, opened 2025-05-07
Add Agent Experiences MCP Server and Improve CID Manager functionality

jfrank-summit:
# Add Auto Experiences MCP Server

This PR adds a new MCP server for managing agent experiences, along with several improvements to the existing codebase.

## Key Changes

1. Added new Auto Experiences server with two main tools:
   - `save-experience`: Saves agent experience data to AutoDrive and updates the last experience CID
   - `retrieve-experience`: Retrieves agent experiences using their CID

2. Enhanced CID Manager functionality:
   - Made contract info optional with graceful fallback to local-only storage
   - Improved error handling and offline mode support
   - Refactored provider initialization logic

3. Infrastructure improvements:
   - Added new binary `auto-experiences-server` for standalone operation
   - Updated main server to support both Auto Drive and Auto Experiences
   - Enhanced error handling and logging across the codebase

4. Type safety improvements:
   - Replaced `any` types with proper TypeScript types
   - Added better type definitions for experience data
   - Improved error handling types

## Testing

- Tested with both Cursor MCP and Claude Desktop clients

### DIFF
--- a/packages/auto-mcp-servers/src/auto-experiences/handlers.ts
+++ b/packages/auto-mcp-servers/src/auto-experiences/handlers.ts
@@ -8,10 +8,50 @@ import {
   ExperienceSaveResult,
 } from '@autonomys/auto-agents/dist/experiences/types.js'
 
-export const createExperienceHandlers = async (options: ExperienceManagerOptions) => {
-  // Initialize ExperienceManager - this handles CidManager and AutoDrive interactions internally
-  const experienceManager = await createExperienceManager(options)
+// Define a type for the ExperienceManager to help with typing
+type ExperienceManager = Awaited<ReturnType<typeof createExperienceManager>>
 
+// Store a singleton instance to ensure we only create one ExperienceManager
+let experienceManagerInstance: ExperienceManager | undefined
+let experienceManagerPromise: Promise<ExperienceManager> | undefined
+
+/**
+ * Ensures that the ExperienceManager is only created once and properly initialized.
+ * @param options Configuration options for the ExperienceManager
+ * @returns A promise that resolves to the ExperienceManager instance
+ */
+const getExperienceManager = async (
+  options: ExperienceManagerOptions,
+): Promise<ExperienceManager> => {
+  // Return existing instance if already created
+  if (experienceManagerInstance) {
+    return experienceManagerInstance
+  }
+
+  // Return in-flight promise if initialization is already in progress
+  if (experienceManagerPromise) {
+    return experienceManagerPromise
+  }
+
+  // Create a new initialization promise
+  console.error('Creating ExperienceManager instance...')
+  experienceManagerPromise = createExperienceManager(options)
+    .then((manager) => {
+      console.error('ExperienceManager created successfully')
+      experienceManagerInstance = manager
+      return manager
+    })
+    .catch((error) => {
+      console.error('Failed to create ExperienceManager:', error)
+      // Clear promise so next call will try again
+      experienceManagerPromise = undefined
+      throw error
+    })
+
+  return experienceManagerPromise
+}
+
+export const createExperienceHandlers = (options: ExperienceManagerOptions) => {
   return {
     /**
      * Saves the provided agent experience data.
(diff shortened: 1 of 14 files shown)


packages/auto-mcp-servers/src/auto-experiences/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  console.error(
+    'EVM environment variables (RPC_URL, CONTRACT_ADDRESS) not fully set. Experience manager will operate in offline mode with local-only CID storage. Wallet private key IS still required for signing uploads.',
+  )
+}
+
+// Create MCP Server
+export const autoExperiencesServer = new McpServer({ name: 'Auto Experiences', version: '0.1.0' })
+
+// Initialize handlers - now fully synchronous with internal async management
+console.error('Initializing Auto Experiences Handlers...')
```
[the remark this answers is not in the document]
jfrank-summit: Local MCP servers should not log messages to stdout (standard out), as this will interfere with protocol operation. [Modelcontextprotocol](https://modelcontextprotocol.io/docs/tools/debugging)
The reason behind this is that the MCP server using the stdio transport communicates with the host application (like Claude Desktop) through the standard I/O streams. The protocol messages are transmitted over stdout, so any additional logging to stdout would corrupt the protocol communication.
Instead, MCP server developers are advised to use stderr for logging:
When building a server that uses the local stdio transport, all messages logged to stderr (standard error) will be captured by the host application (e.g., Claude Desktop) automatically.

See https://modelcontextprotocol.io/docs/tools/debugging#server-side-logging

marc-aurele-besner (approved):
Awesome

packages/auto-mcp-servers/src/bin/auto-experiences.ts
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env node
+
+import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js'
+// Import directly from the root index file
+import { autoExperiencesServer } from '../index.js'
+
```
Xm0onh, line 6: I'm not sure what was the initial design, but if the servers are supposed to be separate, we should import this like below. Otherwise the args of `autoExpereienceServer` will be combined with `auto-drive` MCP servers and will not run.
```suggestion
import { autoExperiencesServer } from '../auto-experiences/index.ts'
```

Xm0onh: For more context, I tried to run the auto-drive MCP and it was giving me error about missing `AGENT_PATH`.

packages/auto-mcp-servers/src/bin/main.ts
```diff
@@ -2,19 +2,32 @@
...
+  For Auto Experiences server:
+    AUTO_DRIVE_API_KEY     API key for Auto Drive (required)
+    NETWORK                'mainnet' (default) or 'taurus'
+    UPLOAD_ENCRYPTION_PASSWORD  Password for encryption (optional)
+    AGENT_PATH             Path to agent (required)
+    AGENT_NAME             Name of agent (required)
+    AGENT_VERSION          Version of agent (optional)
+    PRIVATE_KEY            Wallet private key (required)
+    RPC_URL                EVM RPC URL (optional)
+    CONTRACT_ADDRESS       Contract address (optional)
```
Xm0onh, line 30: Can we likewise add this to the README, as we did for the auto-drive explanation? 😄

Xm0onh (commented):
It's great! Just dropped two comments.

Also in `bin/auto-drive.ts`
I think we should change the import of `autoDriveServer` to this
```
import { autoDriveServer } from '../auto-drive/index.js'
```
Otherwise when you run auto-drive.js, it loads the code from auto-experiences module too because it's imported in the index.ts file and index.ts will come to runtime. The consequence is forcing you to have consodidated args for both servers even if you don't want to use one of them.

packages/auto-mcp-servers/src/bin/auto-experiences.ts
jfrank-summit: good catch

packages/auto-mcp-servers/src/auto-experiences/handlers.ts
```diff
@@ -0,0 +1,110 @@
+import { CallToolResult } from '@modelcontextprotocol/sdk/types.js'
+// Use ExperienceManager for saving/retrieving full experiences
+import { createExperienceManager } from '@autonomys/auto-agents'
+import {
+  AgentExperience,
+  AgentExperienceV0,
+  ExperienceManagerOptions,
```
Xm0onh, line 7: AgentExperienceV0 is marked as deprecated.

Xm0onh (commented):
I just noticed one more thing

packages/auto-mcp-servers/src/auto-experiences/handlers.ts
jfrank-summit: 🤦‍♂️

Xm0onh (approved):
Worked great! 🚀

Xm0onh:
Just one question before merging. The version of MCP is 0.1.2 and the rest of the packages is 0.1.48. Is it intentional?

jfrank-summit:
> Just one question before merging. The version of MCP is `0.1.2` and the rest of the packages is `1.4.18`. Is it intentional?

I read "Worked great!" and hit merge. I should have updated this probably, but I think the version will get updated when all of the packages are updated. @marc-aurele-besner can verify the process.

marc-aurele-besner:
> > Just one question before merging. The version of MCP is `0.1.2` and the rest of the packages is `1.4.18`. Is it intentional?
> 
> I read "Worked great!" and hit merge. I should have updated this probably, but I think the version will get updated when all of the packages are updated. @marc-aurele-besner can verify the process.

Exactly, lerna (a tool to manage mono-repo) at the root of the sdk repo, is set up so when publishing from the root of the repo, all packages get bump to the same version. This allows to know that if the latest version is 0.2.0 as example, all the `@autonomys` packages of that version should be working together, and not have to find individual version of each packages that work with other packages.

marc-aurele-besner:
Basically when running `yarn publish` from the root, it re build everything, then it ask you what type of samver bump you want, will publish a tag to GitHub with this version and store the commit hash as GitHead in each packages.json and publish the new version to npm.

For new package, the package will first need to be publish separatly to npm with `npm publish --access publis` as lerna publish can only bump and not do the initial publish

PR #392, opened 2025-05-09
v1.4.21

clostao:


marc-aurele-besner (dismissed):

iamnamananand996:
Hi @clostao, @marc-aurele-besner is their any specific reason, we update all the packages version, when we make any changes to SDK.


we should consider having a automated release cycle, with the changes log.

cc - @jfrank-summit

marc-aurele-besner:
> Hi @clostao, @marc-aurele-besner is their any specific reason, we update all the packages version, when we make any changes to SDK.
> 
> we should consider having a automated release cycle, with the changes log.
> 
> cc - @jfrank-summit

At the root of the sdk repo, we use lerna to manage the mono repo, by default it keeps all package version align. Personally I find this much more practical vs having specific version for each packages, as it's making it very easy to know what version should work together, when a project import multiple packages.

We don't NEED to publish a new version each time we make changes, but since many of these packages are use by other products we have, often changes are made when bugs or inefficiency are discovered, and therefore I can imagine that the same person that made the changes, want to roll out the new version as soon as the changes has been review and merge.

By "automatic release cycle" if you mean, each time a new PR is merged to main, releasing a version via CI/CD, then we will likely lose the capacity to bump the version with the specific SemVer version bump. If you mean on a specific date cycle, to me this seems like just adding delay and friction to be able to update the product that depend on these package changes.

I think it's pretty easy to see who else is currently working on the SDK, looking at the open or recent PR's and if someone else is also working at the same time on the SDK, you guys could DM's each other to coordinate the release of the new version...

clostao:
> Hi @clostao, @marc-aurele-besner is their any specific reason, we update all the packages version, when we make any changes to SDK.
> 
> we should consider having a automated release cycle, with the changes log.
> 
> cc - @jfrank-summit

I don't have a definite reason for having a common version system. My take is that it would have some drawbacks in terms of inter-compatibility such as knowing what's the package version that fits with other: Imagine you have x.y.z version of `@autonomys/auto-utils`, having a common system would allow to know that the version of `@autonomys/rpc` that matches in time with the installed is x.y.z too. 

As I said it wouldn't be a big problem to not have common version system though I like having common version better

PR #401, opened 2025-05-13
chore: add GITHUB_TOKEN to release workflow environment

clostao:
I think that the error thrown [here](https://github.com/autonomys/auto-sdk/actions/runs/14993109414/job/42120916567) is due to the lack of this env variable.


iamnamananand996 (dismissed):

iamnamananand996 (approved):

clostao:
Removed build and test commands since they're execute in `prepublish` workflow

PR #415, opened 2025-05-27
fix(auto-mcp-servers): remove cross-dependency on env vars

jfrank-summit:
### Problem
The `auto-drive` MCP server was failing to start due to an unwanted dependency on environment variables from `auto-experiences` (`AGENT_PATH`, `AGENT_NAME`, etc.). This occurred because the main entry point was statically importing both servers, causing all modules to be loaded and executed regardless of which server was requested.

### Solution
- Changed static imports to dynamic imports in `src/bin/main.ts`
- Each server module is now only loaded when specifically requested

### Changes
- `packages/auto-mcp-servers/src/bin/main.ts`: Replace static imports with dynamic imports

### Testing
- `auto-drive-server` now starts without requiring `AGENT_PATH` environment variable
- `auto-experiences-server` continues to work as expected with its required environment variables

### DIFF
--- a/packages/auto-mcp-servers/src/bin/main.ts
+++ b/packages/auto-mcp-servers/src/bin/main.ts
@@ -1,8 +1,6 @@
 #!/usr/bin/env node
 
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js'
-import { autoDriveServer } from '../auto-drive/index.js'
-import { autoExperiencesServer } from '../auto-experiences/index.js'
 
 const showHelp = () => {
   console.error(`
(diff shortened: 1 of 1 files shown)


Xm0onh (approved):
LGTM!

P.S. It is interesting, I noticed this error and thought we solved it in PR #386. Wondering what happened. Because it was working at that time but now it's throwing error without this PR fix. 🤔

jfrank-summit:
> LGTM!
> 
> P.S. It is interesting, I noticed this error and thought we solved it in PR #386. Wondering what happened. Because it was working at that time but now it's throwing error without this PR fix. 🤔

We did a partial fix, not sure why it was working at the time or if there was some nuance that just made us think it was working. My first test locally was pointing directly to the auto-drive.ts file in bin, so gave me impression it was working. Possibly something like this happened 🤷‍♂️

PR #430, opened 2025-06-23
fix: downloads should be perfomed in raw mode

clostao:
*Problem:*

Auto-Drive already handles decompression/decryption by its own, so the download should be performed in raw mode.


jfrank-summit (approved):
LGTM

PR #435, opened 2025-06-28
refactor: nominatorPosition calculation into smaller, easier to follow functions

jfrank-summit:
## Changes Made

**🔄 Function Decomposition**

- Broke down the large `nominatorPosition` function into smaller, focused helper functions:
  - `processPendingDeposits()` - handles pending deposit calculations and epoch-based share conversions
  - `processPendingWithdrawals()` - processes withdrawal amounts with proper epoch handling
- Reduced main function complexity from ~150 lines to ~50 lines for better readability

**✨ Removed Mutability **

- Eliminated variable mutations (`totalShares` and `totalStorageFeeDeposit` are now calculated functionally)
- Removed side effects from helper functions - `processPendingDeposits` now returns pending deposits instead of mutating a passed array
- Made all calculations pure and predictable with no external state modifications

**🏗️ Code Quality Enhancements**

- Consolidated duplicate code patterns and logic
- Added proper TypeScript typing using `Deposit` and `Withdrawal` interfaces
- Improved error handling with consistent patterns across helper functions
- Enhanced logging strategy: only log warnings/errors, removed success case noise

**⚡ Enhanced Withdrawal Processing**

- Added epoch-aware withdrawal processing that checks if `domainEpoch[1] < currentEpochIndex`
- Improved share price calculations for withdrawals using historical epoch prices
- Added fallback logic for current epoch withdrawals using operator's current share price

**🧹 Cleanup**

- Removed debug console logs (`depositsData`, `withdrawalsData`, `currentEpochIndex`)
- Streamlined variable naming for consistency (`pendingStorageFee` → `additionalStorageFee`)

## Benefits

- **Maintainability**: Smaller functions are easier to test, debug, and modify independently
- **Predictability**: Functional approach eliminates side effects and makes code behavior more predictable
- **Type Safety**: Proper TypeScript types catch potential runtime errors at compile time
- **Performance**: More efficient epoch-based calculations for withdrawals
- **Readability**: Clear separation of concerns and reduced cognitive load

### DIFF
--- a/packages/auto-consensus/src/position/index.ts
+++ b/packages/auto-consensus/src/position/index.ts
@@ -2,41 +2,160 @@ import type { Api } from '@autonomys/auto-utils'
 import { domainStakingSummary } from '../domain'
 import { deposits, operator, withdrawals } from '../staking'
 import type { NominatorPosition } from '../types/position'
+import type { Deposit, Withdrawal } from '../types/staking'
 import { instantSharePrice, operatorEpochSharePrice } from './price'
 import { shareToStake, stakeToShare } from './utils'
 
 /**
+ * Processes pending deposits to calculate additional shares and storage fees
+ */
+const processPendingDeposits = async (
+  api: Api,
+  operatorId: string | number | bigint,
+  depositData: Deposit,
+  currentEpochIndex: number,
+) => {
+  const zeroAmounts = {
+    additionalShares: BigInt(0),
+    additionalStorageFee: BigInt(0),
+    pendingDeposits: [] as NominatorPosition['pendingDeposits'],
+  }
+
+  if (!depositData.pending) return zeroAmounts
+
+  const {
+    effectiveDomainEpoch,
+    amount,
+    storageFeeDeposit: additionalStorageFee,
+  } = depositData.pending
+
+  // If epoch hasn't passed yet, keep as pending
+  if (effectiveDomainEpoch >= currentEpochIndex) {
+    return {
+      ...zeroAmounts,
+      pendingDeposits: [{ amount, effectiveEpoch: effectiveDomainEpoch }],
+    }
+  }
+
+  // Epoch has passed - convert pending amount to shares
+  try {
+    const epochSharePrice = await operatorEpochSharePrice(api, operatorId, effectiveDomainEpoch, 0)
+    const additionalShares =
+      epochSharePrice !== undefined ? stakeToShare(amount, epochSharePrice) : amount // Fallback: assume 1:1 conversion
+
+    if (epochSharePrice === undefined) {
+      console.warn(
+        `No epoch share price found for epoch ${effectiveDomainEpoch}, using pending amount as is`,
+      )
+    }
+
+    return {
+      additionalShares,
+      additionalStorageFee,
+      pendingDeposits: [],
+    }
+  } catch (error) {
+    console.warn(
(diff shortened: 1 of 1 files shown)


packages/auto-consensus/src/position/index.ts
```diff
@@ -2,41 +2,160 @@ import type { Api } from '@autonomys/auto-utils'
...
+      )
+
+      const withdrawalAmount = sharePrice ? shareToStake(shares, sharePrice) : BigInt(0) // fallback to 0 if no price available
+
+      pendingWithdrawals.push({
+        amount: withdrawalAmount,
+        unlockAtBlock: unlockAtConfirmedDomainBlockNumber,
+      })
+    } else if (domainEpoch[1] >= currentEpochIndex) {
+      const operatorData = await operator(api, operatorId)
```
Xm0onh, line 114: Minor comment: I think the `domainEpoch[1] > currentEpochIndex` condition is not sound, as it essentially means we are ahead of the chain and living in the future LOL

packages/auto-consensus/src/position/index.ts
```diff
@@ -2,41 +2,160 @@ import type { Api } from '@autonomys/auto-utils'
...
+      const lastSharePrice = operatorData.currentTotalStake / operatorData.currentTotalShares
+      const withdrawalAmount = shareToStake(shares, lastSharePrice)
+      pendingWithdrawals.push({
+        amount: withdrawalAmount,
+        unlockAtBlock: unlockAtConfirmedDomainBlockNumber,
+      })
```
Xm0onh: Why not calculate the `instant_share_price`? This would be more accurate.

Xm0onh (commented):
Looks good!
Just dropped two comments

jfrank-summit:
> ### Bug: Deposit Activation Boundary Error
> The `processPendingDeposits` function incorrectly uses the condition `effectiveDomainEpoch >= currentEpochIndex` to determine if a deposit should remain pending. This causes deposits effective in the _current_ epoch (`effectiveDomainEpoch == currentEpochIndex`) to be incorrectly kept as pending instead of being processed and converted to shares, changing the deposit activation boundary. The condition should be `effectiveDomainEpoch > currentEpochIndex`.

It was wrong before.

> ### Bug: Withdrawal Processing Skips Regular Entries
> The `processPendingWithdrawals` function incorrectly skips regular withdrawals. The `continue` statement, intended to handle cases where `withdrawalInShares` is missing, prematurely exits the loop iteration, preventing the processing of `withdrawal.withdrawals` for that entry. This differs from previous behavior where regular withdrawals were always processed.

Intentional

packages/auto-consensus/src/position/index.ts
```diff
@@ -71,104 +185,30 @@ export const nominatorPosition = async (
...
-          effectiveEpoch: effectiveDomainEpoch,
-        })
-      }
-    }
+    // Calculate final totals functionally
+    const totalShares = depositData.known.shares + additionalShares
+    const totalStorageFeeDeposit =
+      depositData.known.storageFeeDeposit +
+      (depositData.pending?.storageFeeDeposit ?? BigInt(0)) +
+      additionalStorageFee
```
Xm0onh, line 199: Are we counting the pending storage fee here twice?

jfrank-summit: Should be fixed

PR #436, opened 2025-06-30
feat: improve type infer

clostao:
**Problem:**

When two notifications are declared in the same rpc definitions the type received as content of notifications is the merge of the two defined notifications content type.

**Solution:**

Avoid types to merge


jfrank-summit (approved):
LGTM

PR #443, opened 2025-07-08
refactor: update sdk structure for avoid deps issues

clostao:
The dependency `stream-fork` was making the auto-drive crashing because had been bundled into `@autonomys/auto-design-system` so couldn't be mocked. 

The utils in file has been moved to `@autonomys/auto-dag-data` where actually is better sat than in `@autonomys/auto-design-system`.

### DIFF
--- a/packages/auto-design-system/src/components/common/FilePreview/FilePreview.stories.tsx
+++ b/packages/auto-design-system/src/components/common/FilePreview/FilePreview.stories.tsx
@@ -1,6 +1,7 @@
 import {
   canDisplayDirectly,
   decryptFileData,
+  FileData,
   needsContentParsing,
   OffchainMetadata,
   processFileData,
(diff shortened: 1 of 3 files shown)


jfrank-summit:
bugbot run

packages/auto-design-system/src/components/common/FilePreview/FilePreview.stories.tsx
```diff
@@ -1,26 +1,17 @@
...
   needsContentParsing,
+  OffchainMetadata,
   processFileData,
-} from '../../../utils'
+} from '@autonomys/auto-dag-data'
+import { NetworkId } from '@autonomys/auto-utils'
+import type { Meta, StoryObj } from '@storybook/react'
+import React, { useCallback, useEffect, useState } from 'react'
+import { fileMock1, fileMock2 } from '../../../mocks/files.mock'
+import type { FileData } from '../../../types'
```
jfrank-summit, line 12: Should this be imported from `auto-dag-data` and removed from `auto-design-system` types? See bugbot [comment](https://github.com/autonomys/auto-sdk/pull/443#pullrequestreview-3001744677).

PR #449, opened 2025-07-14
fix: fix runtime upgrade compatibility for operator epoch share prices

jfrank-summit:
## Description

This PR addresses compatibility issues caused by the runtime upgrade from Perbill (4 bytes, 9 decimals) to Perquintill (8 bytes, 18 decimals) for operator epoch share prices. The standard polkadot-js API queries fail to decode historical Perbill data, leading to errors like "Decoded input doesn't match input" on networks like Taurus with legacy data.

### Key Changes

- **New Storage Utility** (`packages/auto-consensus/src/utils/storage.ts`):
  - Implements raw storage queries to bypass type decoding.
  - Detects format by byte length and scales Perbill to Perquintill (×10^9).
  - Handles undefined/missing data gracefully with warnings.
- **Update Share Price Logic** (`packages/auto-consensus/src/position/price.ts`):
  - Switches from direct API queries to the new utility.
  - Updates inversion formula and docs to use Perquintill.
  - Adds zero-price handling to prevent invalid conversions.
- **Type and Parsing Fixes** (`packages/auto-consensus/src/utils/parse.ts`):
  - Removes obsolete `domainId` from `RawWithdrawal` type (not present in blockchain data).
- **Calculator Script Updates** (`scripts/staking-position-calculator.ts`):
  - Renames parsing function to `parsePerquintill`.
  - Removes redundant tests; aligns with new utility.
- **Export Updates** (`packages/auto-consensus/src/utils/index.ts`): Exports the new utility for reuse.

### Why This Matters

- Fixes decoding errors for historical data on Taurus.
- Maintains backward compatibility without altering core logic.
- Improves robustness for future upgrades (note: Perbill support can be dropped post-Taurus deprecation).

### Testing

- Verified on Taurus with known data (e.g., operator 3, domain 0, epoch 20628): Correctly scales Perbill to Perquintill.
- Additional tests for non-existent keys return `undefined` as expected.
- No regressions in share/stake conversions or position calculations.


Xm0onh (approved):
Lgtm!

PR #453, opened 2025-07-17
feat: add support for byte range cache retrieval

clostao:
Adds support for byte range retrieval in file system caching. It's a prerequisite for byte-range downloads in auto-drive.


jfrank-summit (approved):
LGTM

PR #484, opened 2025-10-20
feat: enhance file system cache stream handling

clostao:
## Why

The issue that this PR is tackling is that if an error happens within the consumer of a stream this will stall the stream due to the reactive approach of Readable streams. 

## What

This PR implements:

* Introduced `createErrorResilientStream` in `streamUtils.ts` to detect stalled streams with a minimal health check API.
* Wraps `fs.createReadStream` calls with this stream manager tool

### DIFF
--- a/packages/utility/file-server/src/caching/index.ts
+++ b/packages/utility/file-server/src/caching/index.ts
@@ -1,3 +1,4 @@
 export * from './defaultConfigs.js'
 export * from './fileCache.js'
+export * from './streamUtils.js'
 export * from './utils.js'


packages/utility/file-server/src/caching/streamUtils.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  // Handle source stream data
+  sourceStream.on('data', (chunk) => {
+    try {
+      metrics.lastActivity = Date.now()
+      metrics.isStalled = false
+
+      if (!passThrough.destroyed) {
+        passThrough.write(chunk)
+      }
```
jfrank-summit: Should you check to see if `passThrough.write(chunk)` succeeds before updating `metrics`? Something like:

```suggestion
const writeSuccess = passThrough.write(chunk)
if (writeSuccess) {
  metrics.lastActivity = Date.now()
  metrics.isStalled = false
}
```

PR #488, opened 2025-10-27
feat: add new wrapper method for upload file as `Buffer`

clostao:
## Description 

Added a simple method for uploading a file as `Buffer`

## Impact

- Simplifies how files are uploaded from nodejs environments

Other:

- Update examples in readme to use Mainnet instead of Taurus as is retired


clostao:
Forgot to push the actual changes 😓

jfrank-summit (approved):
LGTM

Issue #505, opened 2025-12-13
bug: FilePreview component triggers auto-download for unknown file types


0xjeremyfrank:
## Summary

The `FilePreview` component in `@autonomys/auto-design-system` uses an HTML `<object>` tag as a fallback for unrecognized file types. This causes browsers to automatically download files with unknown extensions (e.g., `.enc`, `.bin`, `.dat`) instead of showing a "No Preview Available" message.

## Current Behavior

When a file has an extension that doesn't match any recognized previewable type, the component falls back to rendering:

```jsx
<object
  className="h-[50vh] w-full border dark:border-gray-700"
  data={fileData.uri}
  type={fileData.fileType}
  aria-label={sanitizedFileName}
  title={sanitizedFileName}
/>
```

**Location:** `packages/auto-design-system/src/components/common/FilePreview/FilePreview.tsx` (approximately line 130-140 in the fallback case of `renderFileContent`)

## Problem

When browsers encounter an `<object>` element with:

- `data` pointing to a URL
- `type` set to `application/octet-stream` or another non-renderable type

They cannot display the content inline and instead trigger a file download. This happens automatically when the metadata page loads, without any user interaction.

## Expected Behavior

For file types that cannot be previewed inline, the component should display the `NoPreviewAvailable` component instead of attempting to embed the file with an `<object>` tag.

## Affected File Types

Any extension NOT in the recognized list triggers this issue:

- Images: `jpg`, `jpeg`, `png`, `gif`, `svg`, `webp`, `bmp`, `ico`
- PDF: `pdf`
- Video: `mp4`, `webm`, `avi`, `mov`, `mkv`, `flv`, `wmv`
- Audio: `mp3`, `wav`, `ogg`, `flac`, `m4a`, `aac`
- Text: `js`, `jsx`, `ts`, `tsx`, `html`, `css`, `py`, `java`, `rb`, `go`, `rust`, `php`, `txt`, `md`, `xml`, `csv`
- JSON: `json`

Examples that trigger auto-download: `.enc`, `.bin`, `.dat`, `.exe`, `.zip`, `.tar`, `.gz`, etc.

## Proposed Fix

Replace the `<object>` fallback with `NoPreviewAvailable`:

```tsx
// Current (problematic):
const sanitizedFileName = fileData.fileName ? sanitizeHTML(fileData.fileName) : undefined;
return (
  <div className='flex flex-col items-center'>
    <object
[...]

PR #506, opened 2025-12-17
fix: remove object fallback to prevent auto-downloads (#505)

EmilFattakhov:
This PR:
- Replaces the <object> fallback in FilePreview with NoPreviewAvailable + DirectGatewayLink.
- Prevents browsers from auto-downloading/embedding unknown types (e.g., .enc, .bin, .dat).
- Keeps existing previews for supported types (image/video/audio/PDF/text/JSON) unchanged.
- Removes the now-unused sanitizedFileName variable.


jfrank-summit (approved):
LGTM. I tested locally and it works.

PR #515, opened 2026-02-13
fix: handle wallet rejection and unsubscribe in signAndSendTx

jim-counter:
## Summary

Fix `signAndSendTx` to properly handle wallet signing rejection and clean up subscriptions.

### Problem

When using `signAndSendTx` with browser wallet extensions (SubWallet, Talisman, Polkadot.js), if the user rejects the signing request, the returned promise hangs forever. This is because:

1. `tx.signAndSend()` returns a `Promise<() => void>` that **rejects** when the wallet denies signing, but this rejection was never caught
2. The status callback only fires *after* the transaction is submitted — it never fires if signing is denied
3. The subscription was never unsubscribed, leaking the websocket listener

### Fix

- **Catch wallet rejection:** Attach `.then(fn, safeReject)` to the outer promise so wallet denials propagate to the caller
- **Settled guard:** Prevent redundant processing when the status callback fires multiple times (`isInBlock` then `isFinalized`)
- **Race-safe unsubscribe:** Handle both timing paths — if the callback settles before `.then()` assigns `unsub`, unsubscribe immediately when the outer promise resolves; if `.then()` fires first, store `unsub` for the settlement handler
- **Synchronous throw:** Wrap `signAndSend` in try/catch for wallet extensions that throw synchronously

### Tests

Added 6 unit tests covering:
- Wallet rejection rejects the returned promise
- Slow path: outer resolves before callback — unsub called on settlement
- Fast path: callback settles before outer resolves — unsub called when outer resolves
- Synchronous throw from signAndSend rejects the promise
- Settled guard: `isFinalized` after `isInBlock` is a safe no-op
- Transaction failure (`isDropped`) rejects and unsubs

### Note

`detectTxSuccess` has a pre-existing bug where `return true` inside `forEach` returns from the callback, not the function — so `success` is always `false`. This is unrelated to this PR and does not affect transaction outcomes (the tx still succeeds/fails correctly).

## Test plan

- [x] `npx jest __test__/signAndSendTx.test.ts` — all 6 tests pass
- [x] `npx lerna run build` — all 17 packages build cleanly
- [x] Manual test: Consensus→EVM transfer, reject in SubWallet — spinner stops, error shown
- [x] Manual test: Consensus→EVM transfer, reject in Talisman — spinner stops, error shown

### DIFF
--- a/packages/auto-utils/__test__/signAndSendTx.test.ts
+++ b/packages/auto-utils/__test__/signAndSendTx.test.ts
@@ -1,0 +1,165 @@
+import { signAndSendTx } from '../src/utils/signAndSendTx'
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+/**
+ * Build a minimal mock SubmittableResult.
+ */
+function mockResult(overrides: {
+  status: Record<string, unknown>
+  events?: Array<{ event: { section: string; method: string; data: unknown[] } }>
+  txHash?: string
+}): any {
+  const status: Record<string, unknown> = {
+    isInBlock: false,
+    isFinalized: false,
+    isRetracted: false,
+    isFinalityTimeout: false,
+    isDropped: false,
+    isInvalid: false,
+  }
+  for (const [key, val] of Object.entries(overrides.status)) {
+    status[key] = val
+  }
+  return {
+    status,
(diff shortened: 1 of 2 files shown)


EmilFattakhov (approved):
Looks good!

PR #517, opened 2026-04-07
fix: wallet name misidentification

jim-counter:
## Summary

- **Bug**: When a user connects via Polkadot.js and has no accounts, the error says "No accounts found in **Nova Wallet**" — even without Nova Wallet installed
- **Root cause**: `NovaWallet` and `PolkadotjsWallet` in `@talismn/connect-wallets` both use `extensionName = 'polkadot-js'`. Since `NovaWallet` appears first in the array, `getWalletBySource('polkadot-js')` always returns the Nova entry. `detectWallets` correctly filters Nova out of the UI, but `connectToWallet` did its own independent lookup and hit Nova instead.
- **Fix**: Pass the already-filtered wallet object from `availableWallets` into `connectToWallet`, so the connection uses the same wallet the UI displayed. Falls back to `getWalletBySource` only when `availableWallets` isn't populated (e.g. during rehydration).

## Test plan

- [x] Install Polkadot.js extension with **no accounts**, connect via beneficiary portal → error should say "Polkadot.js" not "Nova Wallet"
- [x] Install Polkadot.js extension with accounts, connect → should work as before
- [x] Test wallet reconnection on page reload (rehydration path)
- [x] Verify Talisman and SubWallet connections still work

### DIFF
--- a/packages/auto-wallet/src/connect.ts
+++ b/packages/auto-wallet/src/connect.ts
@@ -1,21 +1,29 @@
 import { getWalletBySource } from '@talismn/connect-wallets';
+import type { Wallet } from '@talismn/connect-wallets';
 import { address } from '@autonomys/auto-utils';
 import type { WalletConfig } from './types';
 
 /**
  * Core wallet connection logic. Connects to a browser wallet extension,
  * enables it, fetches accounts, and converts addresses to the configured SS58 format.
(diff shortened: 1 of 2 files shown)


EmilFattakhov (approved):
LGTM!

PR #533, opened 2026-05-05
fix: detectTxSuccess always returns false (forEach → some)

### LINKED ISSUE #516: Bug: detectTxSuccess always returns false (forEach return doesn't propagate)
## Summary

`detectTxSuccess` in `@autonomys/auto-utils` always returns `false`, regardless of whether the transaction succeeded. This causes `signAndSendTx` to report `success: false` even for confirmed, in-block transactions.

## Root Cause

In `packages/auto-utils/src/utils/detectTxSuccess.ts`, a `return true` inside `Array.forEach()` returns from the arrow function callback — not from `detectTxSuccess` itself. The function always falls through to `return false`.

```typescript
// Current (broken)
const detectTxSuccess = (events) => {
  events.forEach(({ event: { method, section } }) => {
    if (expectSuccessfulTxEvent.indexOf(`${section}.${method}`) > -1)
      return true   // ← returns from the forEach callback, not detectTxSuccess
  })
  return false       // ← always reached
}
```

## Suggested Fix

Replace `forEach` with `some`:

```typescript
const detectTxSuccess = (events) => {
  return events.some(({ event: { method, section } }) => {
    return expectSuccessfulTxEvent.indexOf(`${section}.${method}`) > -1
  })
}
```

## Impact

Every consumer of `signAndSendTx` that checks `result.success` will always see `false`, even for successful transactions. The transaction itself still lands on-chain — the bug is only in the success reporting.

## Reproduction

```typescript
import { transfer } from '@autonomys/auto-consensus'
import { signAndSendTx } from '@autonomys/auto-utils'

const result = await signAndSendTx(sender, tx)
console.log(result.success)    // always false
console.log(result.txHash)     // valid hash — tx actually succeeded
console.log(result.blockHash)  // valid hash — tx is in a block
```

## Workaround

Infer success from the presence of `txHash` and `blockHash` instead of trusting `result.success`:

```typescript
const success = !!(result.txHash && result.blockHash)
```

## Version

`@autonomys/auto-utils@1.6.9`

jim-counter:
## Summary

`detectTxSuccess` in `@autonomys/auto-utils` always returned `false` because a `return true` inside `Array.forEach()` returns from the arrow-function callback, not from `detectTxSuccess` itself. The function always fell through to `return false`.

Replace `forEach` with `some`, which correctly short-circuits and propagates the boolean result.

## Root Cause

```ts
// Before (broken)
export const detectTxSuccess = (events: EventRecord[]): boolean => {
  events.forEach(({ event: { method, section } }) => {
    if (expectSuccessfulTxEvent.indexOf(`${section}.${method}`) > -1) return true  // ← returns from callback, not detectTxSuccess
  })
  return false  // ← always reached
}

// After (fixed)
export const detectTxSuccess = (events: EventRecord[]): boolean =>
  events.some(({ event: { method, section } }) =>
    expectSuccessfulTxEvent.indexOf(`${section}.${method}`) > -1,
  )
```

## Impact

Every consumer of `signAndSendTx` that checks `result.success` always saw `false`, even for confirmed transactions. The `result.success` field is now correctly `true` when a success event (e.g. `system.ExtrinsicSuccess`) is present.

Closes #516

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/auto-utils/src/utils/detectTxSuccess.ts
+++ b/packages/auto-utils/src/utils/detectTxSuccess.ts
@@ -34,10 +34,7 @@ import { expectSuccessfulTxEvent } from './events'
  *   }
  * })
  */
-export const detectTxSuccess = (events: EventRecord[]): boolean => {
-  events.forEach(({ event: { method, section } }) => {
-    if (expectSuccessfulTxEvent.indexOf(`${section}.${method}`) > -1) return true
-  })
-
-  return false
-}
+export const detectTxSuccess = (events: EventRecord[]): boolean =>
+  events.some(({ event: { method, section } }) =>
+    expectSuccessfulTxEvent.indexOf(`${section}.${method}`) > -1,
+  )


EmilFattakhov (approved):
Looks good!

PR #542, opened 2026-05-29
fix(file-server): don't advertise decompressed Content-Length for com…

### LINKED ISSUE #541: Bug: `handleDownloadResponseHeaders`
### Summary

For files stored compressed-but-not-encrypted (ZLIB/deflate), `handleDownloadResponseHeaders` produces HTTP responses where the advertised `Content-Length` does not match the bytes actually placed on the wire. This causes clients to fail or truncate the transfer — browsers show **"Failed to fetch"** (for `fetch()`/`JSON.parse`) or a **broken image** (for ``), while curl reports content-decoding / partial-file errors. Direct downloads through other code paths work, which is why the issue looks file-specific and intermittent.

### Affected code

`src/http/headers.js` → `handleDownloadResponseHeaders` (the compressed-file branch):

```js
const canUseBrowserDecompression =
  compressedButNotEncrypted && shouldHandleEncoding && !rawMode && !byteRange && !isMediaType;

if (canUseBrowserDecompression) {
  res.set('Content-Encoding', 'deflate');
} else if (compressedButNotEncrypted) {
  shouldDecompressBody = true;
}

// ...
if (shouldDecompressBody) {
  // no Content-Length (chunked) — OK
} else if (byteRange && metadata.size != null) {
  // ...
} else if (metadata.size != null) {
  res.set('Content-Length', metadata.size.toString());   // <-- BUG when Content-Encoding was set
}
```

### Root cause

`metadata.size` is the **decompressed** (original) file size, but for compressed files the body streamed on the wire is the **compressed** (deflate) byte count. There are two distinct failure modes:

**1. Document-navigation path (`Content-Encoding: deflate` is set).**
When `canUseBrowserDecompression` is true, the code sets `Content-Encoding: deflate`, then falls through to the final `else if (metadata.size != null)` and sets `Content-Length: metadata.size`. Per the HTTP spec, `Content-Length` must describe the encoded (on-the-wire, i.e. compressed) length — not the decompressed length. The result is a `Content-Length` / body-length mismatch: clients truncate the compressed stream to `decompressedSize` bytes (or detect the mismatch and abort), so decoding fails.

**2. Preview path — `fetch()` / `` / subresource requests (`shouldDecompressBody = true`).**
Here `isDocumentNavigation(req)` is false, so no `Content-Encoding` is set and the function returns `shouldDecompressBody: true`, signalling that **the caller must decompress the body server-side**. This is a fragile, easy-to-miss contract: any caller that pipes the raw `FileResponse` stream without honoring `shouldDecompressBody` will send raw deflate bytes labeled as `image/png` / `application/json` with no `Content-Encoding`, producing corrupted output. (This is exactly what happens in the `file-retriever` gateway controller today — it ignores the return value.)

### Reproduction

1. Upload a compressed, unencrypted file (e.g. a JSON file or PNG that gets ZLIB-compressed at upload).
2. Request it via the gateway:
   - In a browser preview (`fetch()` or ``) → broken/corrupted content.
   - `curl --compressed https://.../file/<cid>` → content-decoding error (curl exit 61/23) or partial file (exit 18).
3. A plain `curl` without `Accept-Encoding` against the document-navigation path shows `Content-Encoding: deflate` together with `Content-Length: <decompressedSize>`, while the transferred body is the smaller compressed size.

### Expected behavior

The bytes on the wire, the `Content-Encoding` header, and `Content-Length` must be mutually consistent for every path.

### Proposed fix

**A. Never advertise `Content-Length: metadata.size` once `Content-Encoding` is set.** When serving the compressed body verbatim with `Content-Encoding: deflate`, omit `Content-Length` (rely on chunked transfer), since the compressed length is not known at header time:

```js
if (canUseBrowserDecompression) {
  res.set('Content-Encoding', 'deflate');
}
// ...
if (shouldDecompressBody || res.getHeader('Content-Encoding')) {
  // Compressed body on the wire: don't send a decompressed-size Content-Length.
  // Rely on chunked transfer encoding.
} else if (byteRange && metadata.size != null) {
  // ... unchanged
} else if (metadata.size != null) {
  res.set('Content-Length', metadata.size.toString());
}
```

**B. Make the `shouldDecompressBody` contract robust.** Either:
- document it loudly so every caller wires up server-side decompression (pipe through `zlib.createInflate()` when `shouldDecompressBody` is true), **or**
- provide a helper in the package (e.g. `createResponseBodyTransform({ shouldDecompressBody })`) that returns the inflate transform (or pass-through), so callers can't accidentally forget it.

### Secondary note (`Accept-Ranges`)

In the `canUseBrowserDecompression` branch, `shouldDecompressBody` is false so `canAdvertiseRanges` becomes true and `Accept-Ranges: bytes` is advertised with the decompressed `metadata.size`. Byte ranges against a compressed body don't map to decompressed offsets, so ranges shouldn't be advertised whenever the body is compressed on the wire (i.e. when `Content-Encoding` is set), not only when `shouldDecompressBody` is true.

---

### Context on the auto-drive side

For reference, the consuming gateway controller (`auto-files-gateway` → `services/file-retriever/src/http/controllers/file.ts`) currently does:

```js
handleDownloadResponseHeaders(req, res, DownloadMetadataFactory.fromIPLDData(metadata), { byteRange, rawMode })
// ...
pipeline(file.data, res, (err) => { ... })   // pipes raw compressed stream, ignores shouldDecompressBody
```

So once the package side is fixed, the gateway controller should also be updated to honor `shouldDecompressBody` (pipe through `zlib.createInflate()` when true). I can prepare that gateway-side change separately if you want — note that my earlier `fix/gateway-content-length-compressed-files` branch is now obsolete, because this whole area was refactored into `@autonomys/file-server` on the gateway's `main`.

EmilFattakhov:
…pressed bodies

Resolves https://github.com/autonomys/auto-sdk/issues/541

handleDownloadResponseHeaders set Content-Length to metadata.size (the decompressed size) even when sending a compressed body — either verbatim with Content-Encoding: deflate (document navigations) or as raw deflate bytes the caller is expected to inflate. The advertised length then did not match the bytes on the wire, so clients truncated or aborted the transfer (failed fetch, broken images, curl content-decoding errors).

- Omit Content-Length (use chunked transfer) whenever the wire body is compressed: server-side decompression or Content-Encoding: deflate.
- Only advertise Accept-Ranges: bytes when the body is not compressed on the wire and the size is known; ranges don't map to a compressed body.
- Add createResponseBodyTransform() so callers can't silently ignore the shouldDecompressBody contract, and document the contract loudly.

Verified createInflate() (zlib-wrapped, RFC 1950) matches the fflate Zlib output produced by the upload path.

### DIFF
--- a/packages/utility/file-server/src/http/__tests__/headers.test.ts
+++ b/packages/utility/file-server/src/http/__tests__/headers.test.ts
@@ -386,6 +386,32 @@ describe('handleDownloadResponseHeaders', () => {
       expect(result.shouldDecompressBody).toBe(false)
     })
 
+    it('should NOT advertise the decompressed Content-Length when Content-Encoding is set', () => {
+      // Regression: Content-Length must describe the encoded (compressed) length on the
+      // wire, not the decompressed metadata.size. Since the compressed length is unknown
+      // here, omit Content-Length and rely on chunked transfer encoding.
+      const req = createMockReq({ 'sec-fetch-dest': 'document', 'sec-fetch-mode': 'navigate' })
+      const res = createMockRes()
+      const metadata = { ...defaultMetadata, isCompressed: true }
+
+      handleDownloadResponseHeaders(req as any, res as any, metadata, {})
+
+      expect(res.set).toHaveBeenCalledWith('Content-Encoding', 'deflate')
+      const headers = res._getHeaders()
+      expect(headers['content-length']).toBeUndefined()
+    })
+
+    it('should NOT advertise byte ranges when Content-Encoding is set (compressed body on wire)', () => {
+      const req = createMockReq({ 'sec-fetch-dest': 'document', 'sec-fetch-mode': 'navigate' })
+      const res = createMockRes()
+      const metadata = { ...defaultMetadata, isCompressed: true }
+
+      handleDownloadResponseHeaders(req as any, res as any, metadata, {})
+
+      expect(res.set).toHaveBeenCalledWith('Content-Encoding', 'deflate')
+      expect(res.set).toHaveBeenCalledWith('Accept-Ranges', 'none')
+    })
+
     it('should flag decompression for media types even when encoding skipped', () => {
       const req = createMockReq({ 'sec-fetch-dest': 'video', 'sec-fetch-mode': 'no-cors' })
       const res = createMockRes()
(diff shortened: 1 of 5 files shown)


EmilFattakhov:
@bugbot run

jim-counter (approved):
LGTM. I assume this is responsible for the JSON content viewing error on the gateway. Nice detective work!

EmilFattakhov:
Yeah it wasn't super obvious, thanks @jim-counter!

PR #568, opened 2026-06-03
fix(auto-dag-data): validate zlib stream before stamping compression …

### LINKED ISSUE #543: `handleDownloadResponseHeaders` sets `Content-Length` to the plaintext size for encrypted files (truncates ciphertext)
### Summary

For encrypted files, `handleDownloadResponseHeaders` advertises `Content-Length: metadata.size`, but `metadata.size` is the **plaintext/original** size while the body streamed on the wire is the larger **ciphertext** (AES‑256‑GCM adds a one-time salt plus a per-chunk IV and auth tag). The advertised length is therefore smaller than the bytes actually sent, so clients truncate the ciphertext. The truncated stream is missing the trailing GCM auth tag of the final chunk, so client-side decryption fails. The same branch also advertises `Accept-Ranges: bytes` against a body whose byte offsets don't correspond to `metadata.size`.

This is a pre-existing bug, distinct from the compressed‑not‑encrypted `Content-Length` issue fixed in `fix/file-server-compressed-content-length` (commit `2a1d066`). That fix deliberately scoped to compression only.

### Affected code

`packages/utility/file-server/src/http/headers.ts` → `handleDownloadResponseHeaders`. For encrypted files, `compressedButNotEncrypted` is false, so `shouldDecompressBody`/`bodyIsCompressedOnWire` stay false and execution reaches:

```js
const canAdvertiseRanges =
  !shouldDecompressBody && !bodyIsCompressedOnWire && metadata.size != null
res.set('Accept-Ranges', canAdvertiseRanges ? 'bytes' : 'none')   // <-- bytes, against ciphertext
// ...
} else if (metadata.size != null) {
  res.set('Content-Length', metadata.size.toString())             // <-- plaintext size, body is ciphertext
}
```

### Root cause

`metadata.size` (`DownloadMetadata.size`, derived from `data.size` / `totalSize`) is the original plaintext size. The encryption layer (`packages/auto-dag-data/src/encryption/index.ts`) produces a wire body of:

```
32 (salt, once)  +  Σ_chunks [ 16 (IV) + plaintextChunkLen + 16 (GCM tag) ]
```

with `ENCRYPTING_CHUNK_SIZE = 1 MiB`. So overhead = `32 + 32 * ceil(plaintextSize / 1 MiB)` bytes (e.g. **+64 bytes** for a file ≤ 1 MiB). The ciphertext is always larger than `metadata.size`, so `Content-Length` understates the body.

### Reproduction

1. Upload an encrypted file.
2. Request it through the gateway and inspect headers: `Content-Length` equals the plaintext size, but the streamed body is larger (salt + per-chunk IV/tag).
3. The client reads `Content-Length` bytes and stops, dropping the tail; AES‑GCM decryption then fails (auth tag truncated / "unable to authenticate data").

### Expected behavior

`Content-Length`, `Accept-Ranges`, and the bytes on the wire must be mutually consistent for encrypted files.

### Proposed fix

- When the body is encrypted, do **not** advertise `Content-Length: metadata.size`. Either:
  - omit `Content-Length` and use chunked transfer encoding (consistent with the compressed-body fix), **or**
  - compute and advertise the true ciphertext length (`32 + 32 * ceil(size / ENCRYPTING_CHUNK_SIZE) + size`) — but this couples the header layer to encryption-format internals, so chunked is preferred.
- Do **not** advertise `Accept-Ranges: bytes` for encrypted bodies (ranges don't map to plaintext offsets). Treat "encrypted on the wire" like the existing `bodyIsCompressedOnWire` gate.

A clean implementation is to generalize the existing flag, e.g. `bodyDiffersFromMetadataSize = bodyIsCompressedOnWire || shouldDecompressBody || metadata.isEncrypted`, and gate both `Content-Length` and `Accept-Ranges` on it.

### Notes / scope

- Needs test coverage analogous to the compression tests: encrypted file ⇒ no `Content-Length: metadata.size`, `Accept-Ranges: none`.
- Verify the encrypted + `byteRange` path as well (currently it would emit a `206` with a plaintext-based `Content-Range`, which is also wrong).

-

EmilFattakhov:
…flag

---

Resolves #543 

### Problem
`auto-dag-data` stamped `uploadOptions.compression = { algorithm: ZLIB }` onto a file's head node based purely on the caller's claim, without checking that the bytes were actually compressed. Callers that declared `ZLIB` but supplied plaintext (e.g. S3/REST uploads setting a compression header) produced permanently-broken objects: flagged compressed but stored as plaintext, which fails downstream with `ERR_CONTENT_DECODING_FAILED`.

### Fix
- Add `isZlibCompressed()` — a pure, synchronous RFC 1950 header check (exported from the package root).
- In `processBufferToIPLDFormatFromChunks`, validate the first data chunk before building the head node and downgrade the flag to uncompressed when the bytes aren't a valid zlib stream. Applied to both the single-chunk and multi-chunk branches.
- Only runs when encryption is disabled (encrypted bytes are ciphertext and shouldn't be checked), and fails safe — drops the flag rather than throwing.

### Tests
Unit tests for `isZlibCompressed` (including the issue's real PNG/JSON bytes) and integration tests covering single/multi-chunk downgrade, genuine-compression round-trips, and the untouched encrypted path. Full suite: 64/64 passing.


EmilFattakhov:
bugbot run

jim-counter (approved):
LGTM!

PR #621, opened 2026-06-12
chore(ci): align workflow Node versions with engines + extend Renovate scope

jim-counter:
## Summary

Two related fixes for an inconsistency spotted while reviewing #606:

1. **Bump CI Node versions to 20** — three workflows still pinned to Node 18 (`build.yaml`, `release.yaml`, `pr-title-check.yaml`), but `package.json` engines say `>=20.8.0` (going to `>=20.20.2` in #606). Build passes anyway because Yarn 4 treats engines as advisory, but the declaration was misleading.

2. **Extend Renovate's `includePaths`** to include `.github/workflows/*.yaml`. The previous config restricted Renovate to `package.json` files only, so workflow Node versions never got tracked. Also adds a `node.js` group rule so future engines + workflow Node bumps land in one PR.

## Why this matters

Without the Renovate config change, this drift would silently recur — the next time engines are bumped, the workflows would fall out of alignment again. With `.github/workflows/*.yaml` in `includePaths` plus the `node.js` grouping rule, Renovate will catch and group these together.

## Changes

| File | Change |
|---|---|
| `.github/workflows/build.yaml` | `node-version: '18'` → `'20'` |
| `.github/workflows/release.yaml` | `node-version: '18'` → `'20'` |
| `.github/workflows/pr-title-check.yaml` | `node-version: '18'` → `'20'` |
| `.github/renovate.json` | Add `.github/workflows/*.yaml` to `includePaths`; add `node.js` packageRule group |
| `.github/workflows/integration-tests.yaml` | Already `'20'` — no change |

## Test plan

- [x] Renovate config is valid JSON
- [x] CI: `Build all packages and run tests` (now on Node 20)
- [x] CI: `Run integration tests with farmerless dev node`
- [x] CI: `validate` (`pr-title-check.yaml` now on Node 20)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #636, opened 2026-06-29
chore(auto-dag-data): drop @peculiar/webcrypto polyfill, use globalThis.crypto

jim-counter:
## Summary

Removes the `@peculiar/webcrypto` polyfill from `auto-dag-data` and uses the platform-native Web Crypto API (`globalThis.crypto`) instead. The polyfill was necessary when the SDK supported Node versions lacking Web Crypto; the engines floor is now `>=20.20.2` (post-#621), and Node 20+ exposes the same standard Web Crypto API at `globalThis.crypto`.

This also resolves the ESM/CJS interop break that has been blocking Renovate PR #610. `@peculiar/webcrypto` 1.7.x shipped as ESM-only, which the project's Jest config (CJS) cannot load. By removing the dep entirely, the problem becomes moot.

## Changes

| File | Change |
|---|---|
| `packages/auto-dag-data/src/encryption/index.ts` | Import + `crypto` source switched from `@peculiar/webcrypto`'s `new Crypto()` to `globalThis.crypto` |
| `packages/auto-dag-data/package.json` | Dropped `@peculiar/webcrypto` from `dependencies` |
| `packages/auto-dag-data/__test__/encryption.spec.ts` | Added cross-implementation regression test |
| `yarn.lock` | Removed `@peculiar/webcrypto` and `webcrypto-core` |

## Cross-implementation safety

The encryption code in `auto-dag-data` passes all crypto parameters explicitly (PBKDF2 with iterations=100000, hash=SHA-256, key length 256; AES-GCM with explicit IV; salt SHA-256 normalised). No implicit defaults means the Web Crypto spec compels any conforming implementation to produce byte-identical output.

To verify this empirically and protect against future impl swaps, this PR also adds a new test:

```ts
it('decrypts a known fixture (cross-implementation regression)', async () => {
  // Ciphertext produced by @peculiar/webcrypto 1.5.0
  // Must decrypt cleanly under any conforming WebCrypto implementation
  const FIXTURE_HEX = '...'
  // ...
  expect(decryptedBuffer.toString('utf8')).toBe(FIXTURE_PLAINTEXT)
})
```

The fixture was generated using the previous `@peculiar/webcrypto` 1.5.0 implementation. **It still decrypts byte-for-byte under `globalThis.crypto`** — empirical proof of cross-impl compatibility. Any future WebCrypto regression that changes PBKDF2 or AES-GCM output will trip this test.

## Verified

- [x] `yarn workspace @autonomys/auto-dag-data run build` clean
- [x] `yarn workspace @autonomys/auto-dag-data run test` — all **65 tests across 8 suites pass**, including the new fixture regression test
- [x] `@peculiar/webcrypto` and `webcrypto-core` no longer appear in `yarn.lock`
- [x] CI

## Effects on the queue

- **Closes (or supersedes) Renovate PR #610** — the polyfill is gone, the bump is moot.
- One fewer transitive dep, one fewer Renovate PR to triage going forward.

### DIFF
--- a/packages/auto-dag-data/__test__/encryption.spec.ts
+++ b/packages/auto-dag-data/__test__/encryption.spec.ts
@@ -122,4 +122,38 @@ describe('encryption', () => {
       ),
     ).rejects.toThrow('Unsupported encryption algorithm')
   })
+
+  // Regression fixture: ciphertext produced by @peculiar/webcrypto 1.5.0 with
+  // a known plaintext + password. Any WebCrypto implementation (peculiar,
+  // node:crypto.webcrypto, browser) must be able to decrypt this byte-for-byte,
+  // because PBKDF2-SHA256 and AES-GCM are fully specified by the WebCrypto
+  // standard. If this test fails after a crypto-impl swap, encrypted data
+  // produced by previous SDK versions would no longer be readable.
+  it('decrypts a known fixture (cross-implementation regression)', async () => {
+    const FIXTURE_PASSWORD = 'fixture-password-do-not-use-in-real-data'
+    const FIXTURE_PLAINTEXT =
+      'hello world — regression fixture for @peculiar/webcrypto cross-impl compat\n'
+    const FIXTURE_HEX =
+      '983c2e01435eccf98035fd6a24c1614a8a9a6872a879aad0bb3b9aae271674a5' +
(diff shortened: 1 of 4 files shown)


EmilFattakhov (approved):
Looks good!

PR #694, opened 2026-09-14
feat(auto-drive): send requestedBytes in createPaymentIntent and handle credit cap rejection

### LINKED ISSUE #691: Feature: `createPaymentIntent` doesn't send `requestedBytes`, so SDK users skip the server-side credit-cap check
## Summary

`POST /intents` in Auto Drive now accepts an optional `requestedBytes` field and uses it to reject a purchase that cannot fit under the per-user credit cap **before** the user is asked to pay ([autonomys/auto-drive#808](https://github.com/autonomys/auto-drive/pull/808)).

`createPaymentIntent` already receives that number as `sizeBytes`, but sends no request body at all — so every SDK caller keeps the old behaviour: they pay on-chain first and discover the cap afterwards, when the intent lands in `OVER_CAP` and needs an admin to resolve it out of band.

That is the exact failure the backend change exists to remove, and the SDK is the one documented integration path that still hits it.

## Current behaviour

`src/api/calls/payment.ts`:

```ts
export const createPaymentIntent = async (
  api: AutoDriveApiHandler,
  sizeBytes: number,
): Promise<PaymentIntent> => {
  const [contractInfo, intentRes] = await Promise.all([
    getPaymentContractInfo(api),
    api.sendAPIRequest('/intents', { method: 'POST' }),   // <-- no body
  ])
```

`sizeBytes` is used only locally, to multiply the returned rate into `ai3AmountWei`.

The docstring states this as a deliberate fact, and it stops being true once auto-drive#808 ships:

> Note: `sizeBytes` is **not** sent to the Auto Drive API. The POST `/intents`
> endpoint accepts no request body — it returns the current `shannonsPerByte`
> rate. The SDK multiplies that rate by `sizeBytes` to produce `ai3AmountWei`,
> saving the caller from doing the BigInt arithmetic themselves.

## Why it matters

Without `requestedBytes` the backend skips `checkCapHeadroom` entirely — the pre-check is guarded on the field being present. So for an SDK caller:

1. `createPaymentIntent` succeeds and returns an amount to send.
2. The user sends AI3 on-chain. **Irreversible.**
3. The backend confirms the payment, tries to grant credits, and the per-user cap rejects it.
4. The intent goes to `OVER_CAP`, which is terminal. Credits are not granted and the payment is not returned; an admin has to raise the cap and reprocess, or arrange a refund.

With `requestedBytes`, step 1 fails with a `403` naming the cap and the remaining headroom, and nothing is spent.

## Proposed change

Forward the size the function already has:

```diff
   const [contractInfo, intentRes] = await Promise.all([
     getPaymentContractInfo(api),
-    api.sendAPIRequest('/intents', { method: 'POST' }),
+    api.sendAPIRequest(
+      '/intents',
+      {
+        method: 'POST',
+        headers: new Headers({ 'Content-Type': 'application/json' }),
+      },
+      // Decimal string: the endpoint takes the string form as canonical, and
+      // it stays exact past 2^53 where a JSON number would not.
+      JSON.stringify({ requestedBytes: sizeBytes.toString() }),
+    ),
   ])
```

This matches how `watchPaymentTransaction` already posts a body through `sendAPIRequest(relativeUrl, request, body)`.

The endpoint accepts a decimal string, or a JSON number while it is a safe integer. The field is optional and back-compatible — an older Auto Drive deployment ignores it, so the SDK does not need to branch on server version.

### Also worth handling

**1. The error surface.** Once the size is sent, `createPaymentIntent` starts receiving a `403` carrying `{"error":"CREDIT_CAP_EXCEEDED","message":"Purchase of N bytes would exceed the per-user credit cap of M bytes: the account already holds …, leaving … available"}`. Today it renders that as:

```ts
throw new Error(`Failed to create payment intent: ${intentRes.status} ${body}`)
```

i.e. raw JSON inside an Error message. The `message` field is written to be shown to whoever is buying, so it is worth parsing out — and a caller cannot currently distinguish "over cap, ask for less" from any other failure without string-matching. Options: a typed error carrying the code, or at minimum preferring `body.message` when the response body has one.

Note `PaymentIntentStatus` has `OVER_CAP` for the *post*-payment case, but there is no representation of a *pre*-payment cap refusal.

**2. `BigInt(sizeBytes)` has no integer guard.** `sizeBytes` is typed `number`, so `createPaymentIntent(api, 0.5)` throws `RangeError: The number 0.5 cannot be converted to a BigInt because it is not an integer` from the `ai3AmountWei` line — an internal error message for what is really a bad argument. Worth an explicit `Number.isSafeInteger(sizeBytes) && sizeBytes > 0` check with a clear message, especially since the value now also crosses the wire. (Same hazard was just fixed on the auto-drive frontend side.)

**3. Docs.** The `createPaymentIntent` docstring paragraph quoted above, and any integration guide describing the `POST /intents` flow, need updating — `sizeBytes` is now sent, and the call can fail with a cap rejection before any on-chain step.

## Acceptance criteria

- [ ] `createPaymentIntent` sends `requestedBytes` derived from `sizeBytes`.
- [ ] A cap rejection surfaces as something a caller can branch on, not raw JSON in an `Error` message.
- [ ] A non-integer or non-positive `sizeBytes` is rejected with a clear message rather than a `RangeError` from `BigInt`.
- [ ] Docstring no longer claims the endpoint takes no request body.
- [ ] Verified against an Auto Drive build containing autonomys/auto-drive#808: a purchase larger than the remaining headroom fails at `createPaymentIntent` with no on-chain transaction.

## References

- auto-drive PR: https://github.com/autonomys/auto-drive/pull/808
- Affected function: `createPaymentIntent`, in `src/api/calls/payment.ts` as shipped in the package
- Observed in `@autonomys/auto-drive@1.6.14`

yeziR4:
Resolves #691

### Overview
This PR updates `createPaymentIntent` in `@autonomys/auto-drive` according to the specifications in #691:

1. **Input Validation:** Added positive safe integer validation for `sizeBytes`, rejecting non-integers, `<= 0`, `NaN`, and `Infinity` with a clean `TypeError` instead of leaking internal `BigInt` errors.
2. **`requestedBytes` Payload:** Updated `createPaymentIntent` to send `{ requestedBytes: sizeBytes.toString() }` in the `POST /intents` request body so server-side credit-cap checks can be performed prior to on-chain payment.
3. **Structured Error Handling (`CreditCapExceededError`):** Created and exported `CreditCapExceededError` (with code `'CREDIT_CAP_EXCEEDED'`) allowing callers to programmatically catch and branch on over-cap rejections. Parses JSON error response bodies to surface clean error messages rather than raw JSON strings.
4. **Documentation:** Updated docstrings to accurately document the request body and credit-cap error behavior.
5. **Unit Tests:** Added 7 unit tests in `packages/auto-drive/__test__/payment.test.ts` covering validation rules, payload structure, credit cap errors, and general error handling.

### Verification
- `yarn workspace @autonomys/auto-drive build` compiles with 0 errors.
- Jest unit tests pass 7/7 green (`packages/auto-drive/__test__/payment.test.ts`).
- Formatted with Prettier per repository config.


yeziR4:
Hi @EmilFattakhov

I've opened this PR to resolve #691 per your specifications:
- Enforced positive safe integer validation for `sizeBytes`, rejecting non-integers, `<= 0`, `NaN`, and `Infinity` with a clean `TypeError`.
- Updated `createPaymentIntent` to forward `{ requestedBytes: sizeBytes.toString() }` with `Content-Type: application/json` in `POST /intents` so backend credit-cap checks are performed prior to on-chain payment.
- Created and exported `CreditCapExceededError` with code `'CREDIT_CAP_EXCEEDED'` allowing callers to cleanly catch and branch on over-cap rejections, and parsing JSON error messages rather than leaking raw JSON strings.
- Updated docstrings to document the payload and error behavior.
- Added comprehensive unit test suite in `packages/auto-drive/__test__/payment.test.ts` (7/7 tests passing green).

Ready for review whenever you have time. Thank you

yeziR4:
@jim-counter

packages/auto-drive/src/api/calls/payment.ts
```diff
@@ -76,14 +80,37 @@ export const createPaymentIntent = async (
...
+      parsed = JSON.parse(body)
+    } catch {
+      // Body is not JSON
+    }
+
+    if (intentRes.status === 403 && parsed?.error === 'CREDIT_CAP_EXCEEDED') {
+      throw new CreditCapExceededError(parsed.message || 'Credit cap exceeded')
+    }
+
+    const errorMessage = parsed?.message || body
```
EmilFattakhov, line 112: This only covers only covers the coded errors. Auto Drive's `HttpError.handleResponse` sends `{ error: <code>, message } ` when an error carries a machine-readable code, but `{ error: <the message> }` — no message key — for every uncoded one.

packages/auto-drive/src/api/models/payment.ts
```diff
@@ -45,12 +45,7 @@ export type PaymentIntent = {
 
 /** All possible states a payment intent can be in */
 export type PaymentIntentStatus =
-  | 'PENDING'
-  | 'CONFIRMED'
-  | 'COMPLETED'
-  | 'EXPIRED'
-  | 'FAILED'
-  | 'OVER_CAP'
+  'PENDING' | 'CONFIRMED' | 'COMPLETED' | 'EXPIRED' | 'FAILED' | 'OVER_CAP'
```
EmilFattakhov, line 48: Unrelated reformat — please restore the one-member-per-line union.

packages/auto-drive/jest.config.js
```diff
@@ -0,0 +1,13 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
```
EmilFattakhov, line 1: Deviates from convention, what's this module is used for? Please drop it unless there's a clear reason

EmilFattakhov (changes requested):
Thank you for your contribution @yeziR4! A couple of blocking comments that should be relatively easy to address.

EmilFattakhov:
What would a nice addition: since #691 `POST /intents` also gained `paymentMethod (usdc_eth)`, that would be great if you could include this as a part of that PR as well. Optional though, could be addressed separately.

yeziR4:
Thanks for the review @EmilFattakhov! All review comments have been addressed in commit \575ab5eb\:

1. **Uncoded Error Response Handling:** Updated error message extraction to check \parsed?.message || (typeof parsed?.error === 'string' && parsed.error !== 'CREDIT_CAP_EXCEEDED' ? parsed.error : body)\. This seamlessly supports both coded errors (\{ error: <code>, message }\) and uncoded errors (\{ error: <message> }\) from Auto Drive's \HttpError.handleResponse\. Added an explicit unit test for uncoded responses.
2. **Restored Union Formatting:** Restored the one-member-per-line formatting for \PaymentIntentStatus\.
3. **Dropped \jest.config.js\:** Removed \packages/auto-drive/jest.config.js\ and dropped the package-level test script from \package.json\ to adhere to repo conventions.

Ready for review, thank you!

PR #695, opened 2026-09-15
chore(lint): wire up eslint dependencies and add root lint script

### LINKED ISSUE #637: Wire up ESLint properly (currently declared but non-functional)
## Current state

The repo declares ESLint and ESLint plugins in eight `package.json` files and has a working flat config at `eslint.config.mjs`, but **ESLint cannot actually run** in the current state:

```
$ yarn eslint packages/auto-utils/src/
ESLint: 9.39.4

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@typescript-eslint/eslint-plugin'
  imported from /…/eslint.config.mjs
```

`eslint.config.mjs` imports `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser`, but neither is declared as a dependency. The config also calls `FlatCompat` (`@eslint/eslintrc`), `js.configs.recommended` (`@eslint/js`), and `globals.node` (`globals`) — none of which are declared.

Additional findings:
- No `package.json` defines a `lint` script
- No CI workflow runs ESLint (only `commitlint` runs, against PR titles)
- No husky pre-commit hook runs ESLint
- `.vscode/settings.json` and `.vscode/extensions.json` reference `dbaeumer.vscode-eslint`, implying contributors are expected to lint via their IDE — but the broken config means the extension can't load it either

So ESLint is effectively dead weight today: declared, configured, but never executed and silently broken in editors.

## Why fix vs remove

The config exists. The VS Code workspace expects it. The deps are partially declared. The clear original intent was working linting; this is a regression that's gone unnoticed because nothing enforces it.

Removing ESLint entirely is the alternative — also a valid choice — but loses the latent value the config implies.

## What surfaces when ESLint actually runs

After installing the missing peer plugins, running ESLint against `packages/**/src` reports **213 problems (209 errors, 4 warnings)**, broken down:

| Rule | Count | Class |
|---|---|---|
| `@typescript-eslint/no-explicit-any` | 116 | Big — every direct `any` |
| `quotes` | 52 | Auto-fixable |
| `preserve-caught-error` | 25 | Manual — add `{ cause: e }` to re-thrown errors |
| `@typescript-eslint/no-unused-vars` | 7 | Easy cleanup |
| `no-duplicate-imports` | 5 | Easy cleanup |
| `camelcase` | 3 | Easy cleanup |
| `no-useless-assignment` | 2 | Easy cleanup |
| `spaced-comment` | 1 | Auto-fixable |
| `react-hooks/exhaustive-deps` | 1 | Easy cleanup |

**54 of these (53 errors + 1 warning) are auto-fixable via `eslint --fix`.**

### Phase 1 — Plumbing (~20 min)
- Add to root `devDependencies`: `@typescript-eslint/eslint-plugin`, `@typescript-eslint/parser`, `@eslint/js`, `@eslint/eslintrc`, `globals`
- Add a `lint` script to root `package.json`: `eslint "packages/**/src/**/*.{ts,tsx,js,mjs}"`
- Downgrade `@typescript-eslint/no-explicit-any` to `warn` in `eslint.config.mjs` so the 116-occurrence debt is surfaced informationally rather than blocking
- Do **not** add a CI step yet — just make `yarn lint` work locally
- Merging Phase 1 alone fixes the broken-config problem without forcing any code changes

### Phase 2 — Auto-fix (~10 min)
- Run `yarn lint --fix`
- Single mechanical commit, resolves 54 of the 213 problems (mostly `quotes`)

### Phase 3 — `preserve-caught-error` (~30 min)
- Manually fix the 25 sites by wrapping rethrown errors with `{ cause: e }`
- Useful regardless of linting — improves diagnostic stack traces

### Phase 4 — Misc cleanup (~15 min)
- Fix the ~18 remaining easy issues: unused vars, duplicate imports, camelcase, exhaustive-deps, useless-assignment

### Phase 5 — `no-explicit-any` policy
- 116 occurrences. Pick one of:
  - Leave as `warn` (Phase 1 default) — surface in reviews, fix opportunistically
  - Add per-site `// eslint-disable-next-line` annotations where `any` is intentional
  - Refactor to proper types
- This phase is open-ended and can sit indefinitely

### Phase 6 — CI enforcement
- Add a `lint` step to `.github/workflows/build.yaml`
- Only after Phases 2–4 land so the codebase is clean (or close to it)

## Acceptance

- `yarn lint` runs cleanly with `--max-warnings` set appropriately
- ESLint is invoked at least once in CI
- VS Code's ESLint extension successfully loads the config and shows real-time feedback
- Renovate PRs for `eslint`, `eslint-config-prettier`, `eslint-plugin-prettier`, `@typescript-eslint/*` are validating actual functioning lint, not a dead config

## Out of scope

- Replacing Prettier with ESLint's stylistic rules
- Adding new lint rules beyond what's in the existing config
- Mandating any specific resolution for the 116 `any` debt (Phase 5 is a separate decision)

yeziR4:
Part of #637
### Description
Wires up missing ESLint v10 development dependencies and defines the root `yarn lint` and `yarn lint:fix` workspace scripts according to Phase 1 of the linting roadmap (#637).
### Context
Running `yarn lint` at root checks all workspace packages (`packages/*/*/src/**/*.ts`). Maintains existing configurations without breaking CI or forcing source code changes.


jim-counter:
Thanks for taking this on.

Reviewed against the Phase 1 checklist in #637, and the substance is all there:

| Phase 1 item | |
|---|---|
| 5 missing devDeps at root | ✅ all five |
| Root `lint` script with the specified glob | ✅ verbatim |
| `@typescript-eslint/no-explicit-any` → `warn` | ✅ |
| No CI step added | ✅ |
| No source changes forced | ✅ |

`lint:fix` is beyond the spec but it's the Phase 2 workhorse, so I'm happy to keep it.

I did find one blocker and then went a bit further than a pure fix, so I want to explain both rather than just push over your work.

## The blocker: lockfile out of sync

`package.json` declared `"@eslint/js": "^9.21.0"` but `yarn.lock` recorded the root workspace descriptor as exact `npm:9.21.0`. Yarn 4 defaults `enableImmutableInstalls` to true whenever `CI` is set, and our `.yarnrc.yml` doesn't override it, so `yarn install` in `build.yaml` aborts before it ever reaches `yarn build`:

```
➤ YN0028: The lockfile would have been modified by this install, which is explicitly forbidden.
```

This stayed invisible because Actions runs on fork PRs need maintainer approval, so all three workflows sat queued as `action_required` and nothing ever reached `yarn install` - Cursor Bugbot was the only thing reporting, because it runs outside Actions. Easy one to hit: it's what you get if `package.json` is hand-edited after `yarn add`, without a follow-up `yarn install`.

## Why I went to eslint v10 rather than just resyncing

While fixing the pin I checked whether `^9.21.0` was load-bearing, and that turned into a bigger finding.

**The 213-problem baseline in #637 is a v10 number.** On 9.x the same run reports 186. The 27-problem gap is `preserve-caught-error` (25) and `no-useless-assignment` (2), which are in `@eslint/js` v10's `recommended` set (64 rules) and in no 9.x release (61 rules). Majors are when ESLint adds rules to `recommended`. So on 9.x, Phase 3 of the issue would have had nothing to act on, and I'd have written a roadmap against numbers the toolchain couldn't reproduce. That's my error in the issue, not yours.

**v10 also makes the dependency declarations honest.** `eslint@9` shipped `@eslint/js` and `@eslint/eslintrc` as its own dependencies, so `eslint.config.mjs`'s direct imports of both resolved via hoisting even though nothing declared them. That's exactly why the original error in #637 names `@typescript-eslint/eslint-plugin` - it's the *third* import in the file, and the first two were silently resolving as phantom deps. `eslint@10` ships neither, so both declarations become genuinely required, and the tree carries a single copy of each instead of duplicates.

**Nothing is lost by moving now.** I traced the history: no root `lint` script has ever existed, and at `337bbe67` - the commit that introduced the flat config - none of the plugins it imports were declared anywhere in the repo. The config has been unrunnable since the day it was written. There's no green baseline to preserve and no CI job to break, so the only cost of v10 is doing the major bump now instead of in a few months with the same 27 problems surfacing then.

Verified before and after: the config needs **no changes** to run on v10, stderr is completely empty (no deprecation warnings), and all peers are satisfied - `@typescript-eslint` 8.70 declares `^10.0.0` explicitly, `eslint-config-prettier` wants `>=7.0.0`, `eslint-plugin-prettier` wants `>=8.0.0`.

## What I pushed to your branch

```
package.json                    @eslint/js ^10.0.1, eslint ^10.10.0, engines
packages/*/package.json  (7)    eslint ^10.10.0
yarn.lock                       regenerated
```

- **All 8 manifests in lockstep.** Bumping only the root leaves the 7 package-level `eslint` declarations pinning the old tree, so the duplicates survive. They have to move together.
- **`engines` narrowed on the root only**, to `^22.13.0 || >=24` - the intersection of our existing `>=22.0.0` floor and eslint 10's supported range. Not eslint's range verbatim, which would have *lowered* us to Node 20. The three published packages' floors are deliberately untouched: that's a consumer-facing contract (see `fix: keep engines floor at Node 22 while CI moves to 24`), and eslint is a devDependency that never runs in a consumer's install.
- **One targeted `yarn dedupe`** of `@eslint-community/eslint-utils`, which was duplicated at 4.9.1 and 4.10.1. `yarn dedupe --check` now reports nothing eslint-related. The ~320 other dedupe opportunities repo-wide are pre-existing and left to Renovate's `lockFileMaintenance`.

Verified at HEAD: `CI=true yarn install` exits 0, and `yarn lint` runs the config and reports **213 problems (93 errors, 120 warnings)** - matching the issue's baseline exactly.

I've approved the workflow runs, and all three are green on `f438c25c`: Build, Integration Tests, and PR Title Check.

## Could you update the PR description?

Two things in it don't hold, and I'd rather they not end up in the changelog:

**1. `Resolves #637`** - this is Phase 1 of 6. Merging it fixes the broken-config problem, which is real and valuable, but CI enforcement (Phase 6) and the cleanup phases are still open. Could you change it to `Part of #637` or `Refs #637` so the issue stays open on merge?

**2. "Verification - `yarn lint`: Executes cleanly across all packages"** - module resolution is indeed fixed, which I think is what you meant, but `yarn lint` exits 1 with 93 errors. That's completely fine for Phase 1 (the issue explicitly defers cleanup to Phases 2 to 4 and only asks that the config load), but "executes cleanly" reads as green, and the next person along will expect a passing lint. Something like:

> `yarn lint`: the config now loads and runs end to end; no module resolution errors. It exits non-zero with 213 problems (93 errors, 120 warnings), which is the pre-existing debt catalogued in #637 and is addressed in Phases 2 to 4.

Also worth dropping the note about `@eslint/js` being "pinned to `^9.21.0` to match installed ESLint `9.39.4`", since it's moot now.

None of this changes my read of the PR: the Phase 1 work is correct and well scoped, and the plumbing does what the issue asked for. Happy to take it once the description is updated.

yeziR4:
@jim-counter Thanks for the thorough breakdown and for getting the v10 sync and CI workflows green!

I've updated the PR description per your request:
- Changed \Resolves #637\ to \Part of #637\.
- Clarified the verification section regarding \yarn lint\ loading and running end-to-end with the catalogued pre-existing debt of 213 problems (93 errors, 120 warnings) deferred to Phases 2-4.
- Removed the moot note regarding \@eslint/js\ 9.x pinning.

Ready whenever you are to merge, thank you!

jim-counter (approved):
Thanks for the contribution @yeziR4 

cc @EmilFattakhov

PR #696, opened 2026-09-16
feat(file-server): add S3 ListObjectVersions and delete marker helpers

### LINKED ISSUE #690: Feature: move helper functions from Auto Drive S3 into the SDK
As per review: https://github.com/autonomys/auto-drive/pull/789. 

Some functions from [apps/backend/src/core/s3/index.ts](https://github.com/autonomys/auto-drive/pull/789/files/f31e1c249b525433ddc8f42dee2bf6130d27e3d5#diff-1182b65ba45703f296075b932a577aa266ed890cd7b72d132ca76586a984b96b) could be moved to the SDK

yeziR4:
Resolves #690
Refs autonomys/auto-drive#789

### Summary
Following the code review in `autonomys/auto-drive#789` by @jim-counter and tracking issue #690 by @EmilFattakhov, this PR extracts the pure S3 versioning logic and delete marker helpers from `apps/backend/src/core/s3/index.ts` into the SDK's `@autonomys/file-server` package (`src/s3/`):

1. **Delete Marker Helpers**:
   - `deleteMarkerVersionId(deletedAt: Date)`: Formats deterministic delete marker version IDs (`dm-${deletedAt.getTime()}`).
   - `resolveDeleteMarkerResult(deletedAt)`: Formats `DeleteObjectResult` for `DeleteObject` responses with `x-amz-delete-marker` headers.

2. **Types**:
   - `S3VersionEntry`, `S3DeleteMarkerEntry`, `ListObjectVersionsParams`, `ListObjectVersionsResult`, `S3VersionRow`, `DeleteObjectResult`.

3. **Pure Aggregation Engine**:
   - `buildListVersionsResult(rows: S3VersionRow[], maxKeys: number)`:
     - Aggregates version rows by key, marking the first (newest) row of active keys as `isLatest: true`.
     - Synthesises delete markers (`isLatest: true`) for soft-deleted (`pointerDeletedAt`) or owner-removed keys.
     - De-duplicates repeated writes sharing the same CID under a single key.
     - Preserves whole keys across page boundaries, cleanly computing `isTruncated` and `nextKeyMarker`.

### Verification
- Added comprehensive unit tests in `src/s3/__tests__/listObjectVersions.test.ts`.
- All 93 tests in `@autonomys/file-server` pass (`5/5` test suites green).
- Prettier and TypeScript builds pass without errors.


jim-counter:
# Review: auto-sdk #696 - S3 ListObjectVersions + delete marker helpers

Checked against the source: auto-drive#789 head `f31e1c2`, `apps/backend/src/core/s3/index.ts`
and `apps/backend/src/app/controllers/s3/`. Tests 93/93, `tsc --noEmit` clean.

## Verdict

The core choice is right. `buildListVersionsResult` is the valuable pure kernel and the
transcription is faithful; `S3VersionRow` is the correct seam, since auto-drive already imports
`S3ObjectListing` into its repository layer. `deleteMarkerVersionId` deserves sharing because
ListObjectVersions and DeleteObject must agree on the format.

Two problems: the guardrails didn't come with it, and the helpers needed to *use* it were left
behind.

## Blocking

**1. Undocumented over-fetch precondition - silent data loss.** `isTruncated` is set only on
seeing a row from a (maxKeys+1)-th key, so the caller must fetch `maxKeys + 1` distinct keys.
auto-drive does and says so at the call site; the SDK docstring documents only row ordering. A
consumer fetching exactly `maxKeys` keys gets `isTruncated: false` and silently drops the rest
of the bucket. `listObjects.ts` already defends against this exact case with
`computeListObjectsDbLimit` + a `dbLimit` argument ("one extra empty page is harmless, but
silently dropping data is not"). No test covers it.

**2. The two comments that mattered most were dropped.** The `#790` KNOWN LIMITATION (maxKeys
bounds *keys*, not Version+DeleteMarker entries - a deliberate spec deviation, no
`version-id-marker` resume) and the "display-only" caveat on `dm-` ids. The SDK now reads as if
`dm-<epoch>` is a resolvable versionId, when `deleteObjectVersionHandler` unconditionally 403s
and `findVersionByCid` can never match it. These matter more in a library than in the app.

**3. `pointerDeletedAt` / `ownerRemoved` relaxed from required to optional.** Both are required
upstream. Optional removes the only guard against a consumer forgetting to select the
delete/moderation state, and the failure mode is deleted or moderated content reported as a live
latest version. Unlike `md5?`, no rationale is given. Also undocumented: both are read from the
first row only and assumed constant across the key's group.

## Missed

**`planListingEncoding` / `hasXmlIllegalChars` / `encodeS3Key`** (controllers/s3/utils.ts) -
the strongest miss, because it's downstream of this PR: auto-drive's handler calls
`planListingEncoding` on the exact result `buildListVersionsResult` returns. As shipped the SDK
gives you an aggregation engine and withholds the piece that stops its output producing invalid
XML or a 500. Pure, zero-dep, already unit-tested upstream.

**Object Lock / versioning bodies** (`objectRetentionBody`, `objectLockConfigurationBody`,
`bucketVersioningBody`, `objectLegalHoldBody`, `OBJECT_LOCK_RETENTION_YEARS`) - the other half of
"pure S3 versioning logic", encoding an invariant already violated once: bucket DefaultRetention
and per-object RetainUntilDate "were inconsistent before: a 100-year bucket default vs. a
year-9999 per-object date." Exactly what a shared constant prevents recurring.

Lower priority: `parseMultipartParts` (natural companion to the `multipartETag` the SDK already
owns, but needs a `fast-xml-parser` dep - fine to defer); `parseBucketAndKey` / `parseCopySource`.
`sendXML` and `buildObjectLocation` are framework-coupled and correctly left behind.

## Smaller points

- `resolveDeleteMarkerResult` is the only **new** code here, not extracted code - a 4-line ternary
  upstream inlines at two sites. Defensible, but not a headline item, and DeleteObject doesn't
  belong in `listObjectVersions.ts`. Split into `deleteMarker.ts`.
- `ListObjectVersionsParams` is exported but consumed by nothing; `ListObjectVersionsResult` lacks
  the `name`/`prefix`/`maxKeys` its sibling `ListObjectsResult` carries, plus `keyMarker`. Adding
  `finalizeListObjectVersions(params, rows, dbLimit)` to mirror `finalizeListObjects` fixes this
  and blocking item 1 in one move.
- ETag fallback (`md5 ?? cid`) is baked into `S3VersionEntry` but left to the caller for
  `S3ObjectListing`. Extract `objectETag(md5, cid)` into `etag.ts` for both. Related: `formatETag`
  is documented as taking "a raw hex MD5" but is passed a CID.
- The `S3VersionEntry.etag` doc says "matching GET/HEAD/List". It matches List only - GET/HEAD do
  `if (etag) res.set(...)` and omit the header for legacy md5-null objects.

## Test gaps

The 13 tests are well written but cover only happy paths. Missing: the truncation precondition
above; non-adjacent duplicate CIDs (`A, B, A` - the current test uses adjacent rows, which a naive
prev-cid check would also pass); both delete signals set at once; `ownerRemoved` on a multi-version
key; a truncated page ending on a deleted key. The `maxKeys <= 0` test asserts an unresumable page
(`isTruncated: true`, `nextKeyMarker: null`) without saying so - `listObjects.ts` at least explains
that choice in a comment.

## Order of work

Releases here are manual (`workflow_dispatch` + lerna fixed version), so merging doesn't cut a
release and several PRs can publish together. The unit that matters for auto-drive is therefore
the **release**, not the PR.

**Must be in this PR** - blocking 1-3, `finalizeListObjectVersions`, the delete-marker module
split, the test gaps. These are defects in what's being shipped, not added scope. All additive.

**Must be in the same release** - `planListingEncoding` / `hasXmlIllegalChars` / `encodeS3Key`.
The risk is auto-drive getting a version with the aggregation engine but not the encoder its own
handler calls on that engine's output. A separate PR landing before the same release solves that
equally well; author's choice which.

**Separate PRs.** Object Lock bodies (in scope by this PR's own description, small and pure -
worth taking here if offered, but not worth asking for while blocking items are open);
`parseMultipartParts` (adds a `fast-xml-parser` dependency - that decision deserves its own PR);
`objectETag` (refactors the existing `listObjects` path rather than adding to it);
`parseBucketAndKey` / `parseCopySource` (independent, no release coupling).

yeziR4:
Thanks for the thorough and constructive review @jim-counter! All feedback has been addressed in commit `1a66a2f6`:

### 1. Over-fetch Truncation Handling
- Added `computeListObjectVersionsDbLimit(maxKeys)` (`maxKeys + 1`) to fetch an extra distinct key for truncation detection.
- Added `finalizeListObjectVersions(...)` with safety checks: if `distinctKeys.size >= dbLimit`, `isTruncated` is marked `true` and `nextKeyMarker` is populated, ensuring no data is silently dropped.
- Documented the known limitation (#790) regarding key-level vs entry-level pagination and non-resumable `maxKeys <= 0` probes.

### 2. Module Structure & Required Row Fields
- Extracted delete marker logic into `packages/utility/file-server/src/s3/deleteMarker.ts`.
- Documented `dm-<epoch>` display-only synthetic IDs (clarifying that delete markers are not stored content rows and cannot be resolved by `findVersionByCid`).
- Made `pointerDeletedAt: Date | null` and `ownerRemoved: boolean` strictly required on `S3VersionRow` to eliminate implicit false-negative delete/moderation states.
- Added `objectETag(md5?, cid?)` helper in `etag.ts`.

### 3. Pure Helpers Added
- `listingEncoding.ts`: `XML_ILLEGAL_CHARS`, `hasXmlIllegalChars`, `encodeS3Key`, `planListingEncoding`.
- `objectLock.ts`: `OBJECT_LOCK_RETENTION_YEARS = 100`, `bucketVersioningBody`, `objectLockConfigurationBody`, `objectRetentionBody`, `objectLegalHoldBody`.

### 4. Comprehensive Test Coverage
- Non-adjacent duplicate CIDs (`A -> B -> A` collapsed to newest occurrence).
- Both delete signals set simultaneously (`pointerDeletedAt != null` + `ownerRemoved: true`).
- `ownerRemoved` on multi-version keys.
- Truncated page ending on a deleted key.
- `maxKeys <= 0` unresumable page behavior.
- All 8 test suites and 114 unit tests passing cleanly.

packages/utility/file-server/src/s3/__tests__/listObjectVersions.test.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  it('sets isTruncated and nextKeyMarker when storage returned a full batch of distinct keys (distinctKeys >= dbLimit)', () => {
+    const rows = [row('a.txt', 'cid-a'), row('b.txt', 'cid-b'), row('c.txt', 'cid-c')]
+    const params = {
+      bucket: 'test-bucket',
+      prefix: '',
+      keyMarker: null,
+      maxKeys: 2,
+    }
+    const dbLimit = computeListObjectVersionsDbLimit(params.maxKeys) // 3
```
jim-counter, line 355: This test doesn't reach the guard it's named for. With `dbLimit = 3` and 3 distinct keys, `buildListVersionsResult` hits the third key and sets `isTruncated = true` itself, so the `!isTruncated` precondition in `finalizeListObjectVersions` is already false and the block never runs. Both assertions pass on values the inner function produced.

Confirmed by deleting the guard and re-running: 114/114 still green.

The guard itself is correct — it only fires when the caller under-fetches, which is what it's there for. Probed `maxKeys: 2, dbLimit: 2` with two keys and it returns `isTruncated: true, nextKeyMarker: 'b'` as intended. It just has nothing pinning it.

Swapping to an under-fetch reaches the branch:

```ts
const rows = [row('a.txt', 'cid-a'), row('b.txt', 'cid-b')]
const params = { bucket: 'test-bucket', prefix: '', keyMarker: null, maxKeys: 2 }

// caller under-fetched: asked storage for maxKeys keys, not maxKeys + 1
const result = finalizeListObjectVersions(params, rows, 2)

expect(result.isTruncated).toBe(true)
expect(result.nextKeyMarker).toBe('b.txt')
```

Worth keeping the existing case too — it's a valid test of `buildListVersionsResult`'s own truncation path, just under a name that promises something else.

packages/utility/file-server/src/s3/etag.ts
```diff
@@ -3,8 +3,15 @@ import { createHash } from 'crypto'
...
-// Format a raw hex MD5 as a quoted S3 ETag: `"<hex>"`.
-export const formatETag = (hex: string): string => `"${hex}"`
+// Format a raw hex MD5 or CID string as a quoted S3 ETag: `"<value>"`.
+export const formatETag = (value: string): string => `"${value}"`
+
+/**
+ * Format an ETag from either the object's MD5 hash (when present) or fallback CID.
+ * Ensures standard quoted formatting: `"<etag>"`.
+ */
+export const objectETag = (md5?: string | null, cid?: string): string =>
```
jim-counter, line 13: `cid` shouldn't be optional here. It's the mandatory fallback and is never absent — `S3VersionRow.cid` is a required `string`. Making it optional means a one-argument call compiles:

```ts
objectETag(null)               // => "\"\""
objectETag(null, 'bafkLegacy') // => "\"bafkLegacy\""
```

So a consumer writing `objectETag(row.md5)` gets no type error and emits `ETag: ""` for every legacy md5-null object. That's not a valid S3 ETag, and it's a value rclone and the AWS SDK compare for integrity.

I see the empty case is pinned deliberately in `etag.test.ts:35`, so this is a disagreement rather than something missed. My argument against it: the empty string isn't graceful degradation — there's no caller that legitimately has neither an MD5 nor a CID, and the optional parameter is the only thing letting the mistake through.

```ts
export const objectETag = (md5: string | null | undefined, cid: string): string =>
  formatETag(md5 || cid)
```

That drops the `|| ''` branch and the test at :35 along with it, and turns the one-arg slip into a compile error. Worth settling before merge since it's a published export.

(`||` over `??` is right, incidentally — an empty-string md5 isn't a valid md5 and should fall through to the CID.)

jim-counter (changes requested):
Couple of small items to address and I think we're good here. Thanks for the revisions!

yeziR4:
Thanks @jim-counter Both review items have been resolved in `f29c6b29`:

1. **Mandatory `cid` in `objectETag`:** Updated the signature to `objectETag(md5: string | null | undefined, cid: string): string => formatETag(md5 || cid)` so callers cannot omit the fallback CID and accidentally emit an empty string.
2. **Pinned Under-Fetch Guard Test:** Updated the test suite with a dedicated test for `finalizeListObjectVersions` when the caller under-fetches (`maxKeys: 2, dbLimit: 2`), reaching and verifying the safety guard branch.

jim-counter (approved):
LGTM. Thanks for the contribution!

cc @EmilFattakhov

PR #697, opened 2026-09-16
feat(rpc): ease underlying HTTP server setup and fix Zod type resolution

### LINKED ISSUE #356: [@autonomys/rpc] Ease the underlying HTTP server
Currently, setting up the underlying http server in RPC server is a bit too complex and requires improvement in the devEx

```ts
{
      server: {
        httpServer: app.listen(config.port, () => {
          logger.info(`Server is running on port ${config.port}`)
        }),
        callbacks: {},
      },
},
```

### LINKED ISSUE #448: [`@autonomys/rpc`] Type resolution fails when using `zod` validators
When using complex zod validators for defining `ApiMethods` inputs and outputs, an error from typescript is shown:

```
Type instantiation is excessively deep and possibly infinite
```

I've researched a bit about it and seems that composing zod types over a single type (the API definition) could the typescript type resolver to throw this error. 

# Alternatives

* Add a validator field for validating input/output but no inference type is made. So if you wanted to define a method would look like this:

```ts
    unsubscribe_object_mappings: {
      params: defineUnvalidatedType<{subscriptionId: string}>(),
      paramsValidation: z.object({
        subscriptionId: z.string(),
      }),
      returns: defineUnvalidatedType<{success: boolean}>(),
      returnsValidation: z.object({
        success: z.boolean(),
      }),
    },
```

* Substitute for another validation library. There are some lighter libraries for validation that would might allow this kind of type inference.

yeziR4:
Resolves #356
Resolves #448

### Summary

This PR addresses developer experience and typing issues in `@autonomys/rpc`:

1. **Ease Underlying HTTP Server (#356)**:
   - Eliminates redundant internal server creation and unifies HTTP server management.
   - Makes `httpServer`, `callbacks`, and server configuration optional with sensible defaults.
   - Allows passing Express applications / RequestListener functions directly to `httpServer`.
   - Supports passing `port` directly in configuration options or calling `server.listen(port)`.
   - Makes `serverParams` optional in `apiDefinition.createServer`.
   - Exposes `httpServer: http.Server` directly on `WsServer` and `RpcServer`.

2. **Zod Validator Type Resolution (#448)**:
   - Updates `DefinitionTypeOutput` to structurally extract output types from Zod schemas, resolving the `Type instantiation is excessively deep and possibly infinite` error on complex schemas.

3. **Test Suite**:
   - Added comprehensive tests in `__test__/rpc/httpServerDevEx.spec.ts` covering Express app integration, port binding, default instantiation, and complex nested Zod schemas.


EmilFattakhov:
Please tag me for review once bugbot comments have been addressed, thanks!

yeziR4:
@EmilFattakhov All 3 Bugbot comments have been addressed and verified with unit tests in commit \1f117ed2\:

1. **Default server non-RPC request fallback:** Added a 404 Not Found fallback request listener when no custom \httpServer\ or request listener function is provided, preventing non-RPC HTTP requests from hanging connections indefinitely.
2. **Close / listen race handling:** Added state tracking and an asynchronous handler for pending binds. If \close()\ is called before \'listening'\ fires, the server safely terminates and closes connections immediately upon binding without leaking the bound socket or throwing errors.
3. **Overlapping listen calls & port validation:** Added \oundPort\ tracking and pending callback queuing. Calling \listen(samePort, cb)\ while a bind is in progress safely queues \cb\ to execute once listening. Calling \listen\ with a different port while listening or pending throws an explicit error instead of silently misreporting.
4. Added unit tests in \__test__/rpc/httpServerDevEx.spec.ts\ covering all 3 scenarios. All 5 test suites and 28/28 unit tests pass!

Ready for review, thank you!

EmilFattakhov (commented):
Thanks for this, @yeziR4. I checked out the branch and ran build + tests locally (5 suites / 28 tests pass). I also probed edge cases against the built `dist`. Findings below.

### #356: mostly addressed
The issue's original pattern (`httpServer: app.listen(...)`, `callbacks: {}`) still works, and the new zero-config / `port` / request-listener forms work too. A few correctness gaps, detailed inline:

1. **`listen()` failure leaves the server stuck, and `port:` crashes the process on `EADDRINUSE`.** There is no `'error'` handling, so `isListenPending` stays `true` forever and every later `listen(otherPort)` throws `Server listen is already pending`. With the `port` option, `listen` runs inside the constructor and the caller has no hook, so the `'error'` event becomes an uncaught exception. Reproduced locally.
2. **Top-level options silently override nested `server` options with `undefined`.** `createRpcServer({ server: { httpServer: custom }, httpServer: undefined })` drops `custom` and creates a fresh server (reproduced). This is easy to hit with config objects like `httpServer: config.server`.
3. **Options are silently ignored when `server` is a `WsServer`.** `createRpcServer({ server: existingWsServer, port: 8080 })` never listens.
4. **Port check is bypassed for servers the caller already started.** If a server was started outside `listen()`, `boundPort` is `null`, so `listen(otherPort, cb)` calls `cb` as though it bound `otherPort`.

### #448: not demonstrated
I couldn't reproduce `Type instantiation is excessively deep` on **main** with the current setup (zod 3.25.76, `moduleResolution: Node16`). I tried the new spec's schemas, 60 methods using `extend`/`merge`/`discriminatedUnion`/`refine`/`transform`, recursive `z.lazy`, a consumer with a *different* zod copy (3.23.8), and legacy `moduleResolution: node`. Main and this branch produce the same results with nearly identical instantiation counts (~72k vs ~72k on the new spec). The new "Issue #448" test is a runtime test that also passes unchanged against main's `typing.ts`, so it doesn't guard the fix.

The `{ _output: infer O }` change is harmless and arguably cleaner, but unless there is a reproduction that fails on main, please change `Resolves #448` to `Refs #448`. That way the issue stays open for the real fix: zod as a `peerDependency` per https://zod.dev/library-authors, and/or the zod 4 upgrade mentioned in the issue thread. If you have a failing repro, please add it as a type-level test (for example a `.ts` fixture compiled with `tsc --noEmit`, or a `expectTypeOf` assertion) that fails on main.

### Other
- `prettier --check` fails on `src/ws/server.ts` and `src/rpc/api/typing.ts`. Please run `yarn format`.
- `WsServer` and `RpcServer` gain a required `httpServer` member. That's fine for a minor release, but it is technically a type break for anyone who hand-implements these interfaces (for example mocks passed as `server`). Worth noting in the PR description.
- Tests: fixed ports 19881–19887 plus manual `close()` at the end of each test means a failed assertion leaks a listening server into later tests. Consider `listen(0)` + `httpServer.address().port` and cleanup in `afterEach`.

packages/utility/rpc/src/ws/server.ts
```diff
@@ -72,18 +89,69 @@ export const createWsServer = ({
...
+        httpServer.closeAllConnections?.()
+        return
+      }
+      while (pendingListenCallbacks.length > 0) {
+        const callback = pendingListenCallbacks.shift()
+        callback?.()
+      }
+    })
+
+    httpServer.listen(listenPort)
```
EmilFattakhov: **Stuck state / uncaught crash on bind failure.** If `httpServer.listen` emits `'error'` (for example `EADDRINUSE` or `EACCES`), the `once('listening')` handler never fires. `isListenPending` stays `true`, the pending callbacks leak, and every later `listen()` throws `Server listen is already pending on port X`, so the server can't be retried. With no `'error'` listener, Node also throws it as an uncaught exception. That is especially bad for the `port` option, where the caller gets no chance to attach a handler (reproduced: `createWsServer({ port: <busy port> })` gives `uncaughtException EADDRINUSE`).

Suggestion: register a paired `once('error')` that resets state and surfaces the error, and remove both listeners on whichever event fires first:

```ts
const onError = (err: Error) => {
  httpServer.off('listening', onListening)
  isListenPending = false
  boundPort = null
  pendingListenCallbacks.length = 0
  onListenError?.(err) // or: rethrow if no handler was provided, to keep today's semantics explicit
}
httpServer.once('error', onError)
```

Also consider accepting a callback for the `port:` form (for example `onListening`) or returning a promise. Right now the only way to know when auto-listen finished is to poke at `httpServer.listening` / `once('listening')`, which is what the tests do.

packages/utility/rpc/src/ws/server.ts
```diff
@@ -72,18 +89,69 @@ export const createWsServer = ({
...
+    }
+  }
+
+  const listen = (listenPort: number, cb?: () => void) => {
+    if (isClosed) {
+      throw new Error('Cannot listen on a closed server')
+    }
+
+    if (httpServer.listening) {
+      if (boundPort !== null && boundPort !== listenPort) {
```
EmilFattakhov: When the caller passes an `http.Server` they already started (for example `app.listen(3000)`, which is exactly the pattern from #356), `boundPort` is still `null` here. `listen(9999, cb)` then calls `cb()` as if it bound 9999 (reproduced). Consider reading the actual port from `httpServer.address()` when `boundPort` is `null`, and compare against that.

packages/utility/rpc/src/ws/server.ts
```diff
@@ -72,18 +89,69 @@ export const createWsServer = ({
...
+      }
+    })
+
+    httpServer.listen(listenPort)
   }
 
-  const listen = (port: number, cb?: () => void) => {
-    internalHttpServer.listen(port, cb)
+  if (typeof port === 'number') {
+    listen(port)
```
EmilFattakhov: Nit: `listen(port)` binds on all interfaces. That's unchanged behaviour, but now that `port` is a first-class option, a `host?: string` option (passed through to `httpServer.listen(port, host)`) would let people bind to `127.0.0.1` for local-only RPC without dropping back to a custom `http.Server`. That is a sensible default-safe knob for an RPC endpoint.

packages/utility/rpc/src/ws/server.ts
```diff
@@ -1,31 +1,22 @@
...
-    autoAcceptConnections: false,
-  })
-
-  if (onClose) {
-    ws.on('close', onClose)
-  }
+import { CreateWsServerParams, WsMessageCallback, WsServer } from './types'
+
+const isHttpServer = (s: unknown): s is http.Server =>
+  s instanceof http.Server ||
```
EmilFattakhov: The duck-typed fallback (`'listeners' in s && typeof listen === 'function'`) accepts any `EventEmitter` with a `listen` method, for example a raw `net.Server`, which never emits `'request'`/`'upgrade'`. I assume the fallback exists for `https.Server`, which isn't an `http.Server` subclass. `s instanceof net.Server` covers http, https, and http2, and is precise. If you want to keep the duck typing for cross-realm cases, at least check for `'request'` semantics (for example `typeof (s as any).setTimeout === 'function' && 'maxHeadersCount' in s`).

packages/utility/rpc/src/rpc/server.ts
```diff
@@ -16,20 +17,21 @@ import {
...
-}): RpcServer => {
-  const wsServer = isWsServer(server) ? server : createWsServer(server)
+export const createRpcServer = (params: CreateRpcServerParams = {}): RpcServer => {
+  const { server, initialHandlers, ...restOptions } = params
+
+  let wsServer: WsServer
+  if (isWsServer(server)) {
+    wsServer = server
+  } else if (typeof server === 'object' && server !== null) {
+    wsServer = createWsServer({ ...server, ...restOptions })
```
EmilFattakhov: `{ ...server, ...restOptions }`: an explicitly `undefined` top-level key overrides the nested value. `createRpcServer({ server: { httpServer: custom }, httpServer: undefined })` silently drops `custom` and creates a fresh server (reproduced). This is common with config objects (`httpServer: config.httpServer`).

Also, having two ways to pass the same options (`server: {...}` and top-level) invites confusion about precedence. Options:
- merge only defined keys (`Object.fromEntries(Object.entries(restOptions).filter(([, v]) => v !== undefined))`), or
- (preferred) type it as a union so the two forms are mutually exclusive, with `server` kept only for the `WsServer` case and the legacy nested form, and the flat options as the new path.

packages/utility/rpc/src/rpc/server.ts
```diff
@@ -16,20 +17,21 @@ import {
...
-  server: WsServer | Parameters<typeof createWsServer>[0]
-  initialHandlers?: RpcHandlerList
-}): RpcServer => {
-  const wsServer = isWsServer(server) ? server : createWsServer(server)
+export const createRpcServer = (params: CreateRpcServerParams = {}): RpcServer => {
+  const { server, initialHandlers, ...restOptions } = params
+
+  let wsServer: WsServer
+  if (isWsServer(server)) {
+    wsServer = server
```
EmilFattakhov: When `server` is an existing `WsServer`, any top-level `port` / `httpServer` / `callbacks` / `onConnection` is silently ignored (`createRpcServer({ server: ws, port: 8080 })` never listens). Either reject that combination at the type level (discriminated union) or throw/warn at runtime.

packages/utility/rpc/src/rpc/api/typing.ts
```diff
@@ -25,7 +25,11 @@ export type MessageDefinition = {
 export type DefinitionType = ZodType | UnvalidatedType<any>
 
 export type DefinitionTypeOutput<T extends DefinitionType> =
-  T extends ZodType<any> ? T['_output'] : T extends UnvalidatedType<infer U> ? U : never
+  T extends { _output: infer O }
```
EmilFattakhov: I couldn't get main to produce `Type instantiation is excessively deep` with the current zod (3.25.76) + `Node16` resolution. That includes deeply composed schemas, recursive `z.lazy`, a consumer using a different zod copy, and legacy `node` resolution. Main and this change behave the same, with the same instantiation counts. So I don't think this line is what resolves #448 (it may have been fixed by the zod bump / tsconfig change the Zod author mentioned in the issue).

The change is fine to keep, but please either add a type-level repro that fails on main, or switch the PR to `Refs #448`.

packages/utility/rpc/__test__/rpc/httpServerDevEx.spec.ts
```diff
@@ -0,0 +1,372 @@
...
+      // Calling listen with a different port once already listening throws
+      expect(() => {
+        server.listen(TEST_DEVEX_PORT_7 + 10)
+      }).toThrow()
+
+      server.close()
+    })
+  })
+
+  describe('Issue #448: Zod validator type resolution', () => {
```
EmilFattakhov: This block passes unchanged against main's `typing.ts` (I swapped it in and ran `tsc --noEmit` + jest), so it doesn't guard the typing change. A regression test for a type-checker error needs to be a compile-time check that fails before the fix.

packages/utility/rpc/__test__/rpc/httpServerDevEx.spec.ts
```diff
@@ -0,0 +1,372 @@
...
+import { z } from 'zod'
+import {
+  createApiDefinition,
+  createRpcClient,
+  createRpcServer,
+  createWsServer,
+  defineUnvalidatedType,
+} from '../../src'
+
+const TEST_DEVEX_PORT = 19881
```
EmilFattakhov: Nit: hard-coded ports plus `close()` at the end of each test body means one failed `expect` leaves a server bound, which can cascade into `EADDRINUSE` in later tests (and, because of the missing `'error'` handling above, into an uncaught exception). Prefer `listen(0)` + `httpServer.address().port`, and close in `afterEach`/`finally`.

